Issue #3, opened 2016-09-09
Create location model


ricco386:
Create model file:

title: CharField, max=128
street: CharField, max=128
street2: CharField, max=128
city: CharField, max=128
postcode: CharField, max=12
state: CharField, max=128
capacity: IntegerField

Once finished, please cross the issue off in list stored at #1 Create initial DB models and migrations

Benjji:
Robime na tejto issue - vlavo od vchodu do miestnosti - nesiahat! :)

jangondol:
Resolved by #25

PR #25, opened 2016-09-13
Location model

Benjji:
Vytvoreny location model #3


jangondol:
There seem to be several [PEP8](https://www.python.org/dev/peps/pep-0008/) issues. Merging this but please clean up later if possible.

PR #33, opened 2016-09-13
Discount Codes model

TomasHo:


ricco386:
Unfortunately there has been another merge request done before this which caused conflict. Please pull the changes from upstream and fix the conflict.

If you leave your code at the end of the file, leave last line blank, so the code is according to [PEP8](https://www.python.org/dev/peps/pep-0008)

TomasHo:
Issues fixed, discount types added.

ricco386:
@TomasHo did you commit correct version? it seems you have removed the class at all, commit 6ea846526d is removing whole class DiscountCodes(models.Model)

TomasHo:
Sorry, should be ok now.

ricco386:
@TomasHo sure thanks, merging the code into upstream.

For the future at the end of the file, leave last line blank, so the code is according to [PEP8](https://www.python.org/dev/peps/pep-0008)

PR #39, opened 2016-09-14
Talk model

mimmon:
Talk model. Added also lightning talk choice, just in case. If unwanted, please create an issue, it will be removed.


konfera/models/talk.py
```diff
@@ -1,5 +1,36 @@
...
 
 class Talk(models.Model):
-    pass
+    title = models.CharField(max_length=256)
+    abstract = models.TextField()
+    talktype = models.CharField(max_length=32, choices=TALK_TYPE_CHOICES)
+    status = models.CharField(max_length=32, choices=TALK_STATUS_CHOICES)
+    duration = models.IntegerField(choices=DURATION_CHOICES)
+    primary_speaker = models.ForeignKey(Speaker)
+    secondary_speaker = models.ForeignKey(Speaker)
```
ZoltanOnody: `secondary_speaker` should be optional.

konfera/models/talk.py
```diff
@@ -1,5 +1,36 @@
...
+    event = models.ForeignKey(Event, blank=True)
```
ricco386: All talks should reference some event, so please remove: blank=True

ricco386:
@mimmon there was already MR by @TomasHo who added talk model. However it didnt had the idea to have a lightning talk as a choice, which I like, so if you could update from upstream and amend the model. Thanks.

mimmon:
closing pull request, sent a new one

Issue #15, opened 2016-09-09
Create list sponsors view


ricco386:
Create simple view (with template) that will list all sponsors per event and according their level and alphabetically in the DB.

TomasHo:
working on this

PR #43, opened 2016-09-15
#15 sponsor list view + basic templates

TomasHo:
sponsor_list_view added, basic templates created


konfera/views.py
```diff
@@ -1,3 +1,14 @@
 from django.shortcuts import render
+from konfera.models import Event, Sponsor
 
-# Create your views here.
+
+def sponsor_list_view(request):
+    events = Event.objects.all()
+    sponsors = Sponsor.objects.all()
```
ricco386: This is not very effective, you are doing select for all sponsors and all events. This would be much more effective since if we have million event and 10 sponsors per event, you will do two DB queries that will result in 11 million results, which you will later order and cycle through.

May be I wasn't clear enough in the issue, it should be all sponsors per event. So in url you define event, and in the view [.get](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#get) it and than just [.filter](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#filter) the sponsors for that event. This way you will do DB queries that will result in 11 results, which you will later order and cycle through.

And if you check out [.select_related](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#select-related) Django will construct more complicated SQL query that will follow foreign keys and you will have only DB query and in the template you will have only one for cycle.

TomasHo:
Know issue - sponsor sorting based on their type name, not type priority

konfera/views.py
```diff
@@ -1,14 +1,15 @@
...
 
 
-def sponsor_list_view(request):
-    events = Event.objects.all()
-    sponsors = Sponsor.objects.all()
-    context = {'events': events,
+def sponsor_list_view(request, event_id):
+    sponsors = Sponsor.objects.select_related('event') \
+                    .filter(event__id=event_id).order_by('type')
+    event = Event.objects.get(pk=event_id)
```
ricco386: to have ids from DB in urls is bad practice. Lets introduce slug field in event model. And update urls to contain this file, and update view logic search by slug field.

This way we give user ability to define his own url per event which is nice.

TomasHo: good point, will do this.

konfera/views.py
```diff
@@ -1,14 +1,15 @@
...
 from konfera.models import Event, Sponsor
 
 
-def sponsor_list_view(request):
-    events = Event.objects.all()
-    sponsors = Sponsor.objects.all()
-    context = {'events': events,
+def sponsor_list_view(request, event_id):
+    sponsors = Sponsor.objects.select_related('event') \
+                    .filter(event__id=event_id).order_by('type')
```
ricco386: How about you do the first event and than you select related sponsors? This way we could get rid of one DB query. Please test if it works this way.

konfera/urls.py
```diff
@@ -0,0 +1,23 @@
...
+    1. Import the include() function: from django.conf.urls import url, include
+    2. Add a URL to urlpatterns:  url(r'^blog/', include('blog.urls'))
+"""
+from django.conf.urls import url
+from django.contrib import admin
+from konfera import views
+
+urlpatterns = [
+    url(r'^event/(?P<event_slug>[\w,-]+)/$', views.sponsor_list_view, name='event_sponsors'),
+    url(r'^admin/', admin.site.urls),
```
ZoltanOnody: admin by tu nemal byt, ten sa bude importovat v projekte

konfera/views.py
```diff
@@ -2,14 +2,15 @@
...
     sponsors = Sponsor.objects.select_related('event') \
-                    .filter(event__id=event_id).order_by('type')
-    event = Event.objects.get(pk=event_id)
+                    .filter(event=event).order_by('type')
+
     context = {'event': event,
-               'sponsors': sponsors,
-               }
+               'sponsors': sponsors, }
+    template = 'konfera/event_sponsors.html'
```
ricco386: I dont think it is necessary to introduce new variable here.

It not used anywhere else in the view, so I think original setup (pass string as a parameter or render) was ok, please revert back.

konfera/admin.py
```diff
@@ -4,10 +4,15 @@
                             TicketType, DiscountCodes, Ticket, Speaker,
                             Talk, Room, Schedule)
 
+
+class EventAdmin(admin.ModelAdmin):
+    prepopulated_fields = {"slug": ("title",)}
```
ricco386: :+1:

ricco386:
Everything seems fine, but there is conflict with the upstream, could you pull from the upstream and resolve the conflict?
If you want you can also regenerate the migration (remove the old one and crete new one, it will change only date and add your new field in models). We can do this since it is not deployed anywhere and this is the iniital one. Once 0.1 is released we will do additional migrations 0002, 0003 etc.

Thanks.

konfera/models/event.py
```diff
@@ -1,5 +1,5 @@
 from django.db import models
-
+from konfera.models import Location
```
ricco386: There is some problem with import (desnt seems to be working), please remove this import as it breaks the automated tests: https://travis-ci.org/pyconsk/django-konfera/builds/160476772 

If you do `location = models.ForeignKey('Location')` (notice that 'Location' is string), Django will handle it and you dont need import.

konfera/views.py
```diff
@@ -1,3 +1,16 @@
-from django.shortcuts import render
+from django.shortcuts import render, get_object_or_404
+from konfera.models import Event, Sponsor
```
ricco386: Please check the result of automated test, if this is not the same problem as mentioned above? Or you can test locally if you run server works?

If it will be a problem you can allways import like this:
`from konfera.models.event import Event` and it will work...

ricco386:
well done.

I am merging into master. If get bored you can try to write some unit tests...

Issue #1, opened 2016-09-09
Create initial DB models and migrations


ricco386:
All models should have, created, modified datetime fields, that are updated automatically. 
If there is status, it should have status_modified field updated automatically. 

List of issues to be done in order to have initial DB schema

Tables that doesnt have FK, so they can be created first:

- [x] #3 Create location model
- [x] #13 Create receipt model
- [x] #8 Create room model

****

Models required for event management, should be created afterwards:

- [x] #2 Create events model
- [x] #4 Create sponsor model 
- [x] #5 Create speaker model 
- [x] #6 Create talk model
- [x] #7 Create schedule model

****

Models for receipts and billing, should be create lastly

- [x] #9 Create tycket_type model
- [x] #12 Create order model
- [x] #11 Create discount_codes model
- [x] #10 Create Ticket (Atendee) model

****

Once all done create initial migration.

- [x] generate initial migration

Django Documentation: https://docs.djangoproject.com/en/1.10/topics/db/models/

ZoltanOnody:
working on it

PR #44, opened 2016-09-15
generate initial migrations, close issue #1

ZoltanOnody:
I had some `import` problems with making migrations so I changed the `ForeinKey`s to be strings. 

Before merging it, I have a question. In most of `ForeinKey`s there is `null=False` which is default in [django](https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L141). Personally I'd delete it, what do you say?


konfera/models/speaker.py
```diff
@@ -11,7 +10,7 @@ class Speaker(models.Model):
     url = models.URLField()
     social_url = models.URLField()
     country = models.CharField(max_length=64)
-    sponsor = models.ForeignKey(Sponsor, default=None)
+    sponsor = models.ForeignKey('Sponsor', default=None)
 
     def __str__(self):
         return self.title
```
ricco386: This is an error from, that I have missed in review. There is no title in this model, however it might be nice to have optional title field for user with choices (Mr., Ms.). Could you please update function to return title (if set) first_name last_name?

I realize it is not part of your issue, but it would be nice to have it before migration, same goes for ticket model.

ZoltanOnody:
**I've done the changes you suggested.**

One of the tests is failing, I am not sure what should be the correct behavior.
```
..........F.
======================================================================
FAIL: test_string_representation (konfera.tests.test_models.TicketTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/zoli/projects/django-konfera/konfera/tests/test_models.py", line 86, in test_string_representation
    '%s %s' % (entry.first_name, entry.last_name))
AssertionError: 'Ticket of Mr/Mrs/Ms Tester' != 'Test Tester'
- Ticket of Mr/Mrs/Ms Tester
+ Test Tester
```
Currently the `Ticket` doesn't have `title`, so either we add it to the model or we remove it from the test.

konfera/models/speaker.py
ZoltanOnody: done - https://github.com/pyconsk/django-konfera/pull/44/commits/3a42915f9ffe262016a00aa95b1471a86be8a87d

ricco386:
@ZoltanOnody thanks, looks great. 

I will fix the last failing test, no worries... Ticket, also represents attendee (as formally it is ticket but it represents person) as each conference attendee will have to have his own ticket.

PR #48, opened 2016-09-16
Fixed failing test

ricco386:
I have updated ticket model and regenearete migration in order to have new field.
Updated test so they do not fail and all of them works.


ricco386:
from the error messages we can [support python 3.3 only for LTS Django](https://docs.djangoproject.com/en/1.9/faq/install/#what-python-version-can-i-use-with-django) I am going to update travis, so we have nice green badge

ZoltanOnody:
:+1:

PR #51, opened 2016-09-16
run flake8 before_script

ZoltanOnody:
we are expecting failure

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,10 @@ matrix:
     - python: "3.3"
       env: DJANGO_VERSION=1.8.14
 
-    # Flake 8
     - python: "3.5"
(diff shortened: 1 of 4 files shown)


ricco386:
or better use before_install (this might not even install if there is problem with pep8) instead of before_script.

Or we can update existing include scrip and test pep8 as part of LTS test, but I guess if we set extra env: TEST_PEP8=True it would be clear directly from the test...

ZoltanOnody:
It seems to be working but I'll try one more thing

ricco386:
seems working, however I am wondering if it will also work if the PEP8 is ok? Just curious if it doesnt crash during the Django install since it is not specified in envs...?

ZoltanOnody:
It seems to be working :)

Issue #19, opened 2016-09-10
Create list events view


ricco386:
Create simple view (with template) that will list all events in the DB order by upcoming date.

Redpoint1:
still not closed?

ricco386:
sorry for delay...

PR #57, opened 2016-09-16
list events from DB(#19)

kapucko:
first try


konfera/urls.py
```diff
@@ -18,4 +18,5 @@
 
 urlpatterns = [
     url(r'^event/(?P<event_slug>[\w,-]+)/$', views.sponsor_list_view, name='event_sponsors'),
+    url(r'^event/$', views.event_list, name='events')
```
ZoltanOnody: IMO add a comma to the end of the line so when someone adds a new `url` git doesn't show 3 changed instead of 1.

konfera/views.py
```diff
@@ -13,3 +14,20 @@ def sponsor_list_view(request, event_slug):
     return render(request=request,
                   template_name='konfera/event_sponsors.html',
                   context=context, )
+
+
+def event_list(request):
+    events = Event.objects.all().order_by('date_from')
+    paginator = Paginator(events, 2)
```
ZoltanOnody: 2 is good for testing but I think we should show more of them on a single page (of course it depends on the design which we don't have for now)

konfera/urls.py
ricco386: :+1:

konfera/views.py
ricco386: I agree lets start with 5 which is also fairly good for testing ;)

Issue #53, opened 2016-09-16
Minor issues in models


ZoltanOnody:
I went through the app and find the following bugs/limitations. 

## Location
- [x] `street2` is required. Does it need to be?

## Events
- [x] we need to enforce `date_from` to be earlier than `date_to`

## Discount codes
- [x] ~~it shows **Discount codess** in admin. `verbose_name_plural` must be changed.~~
- [x] rename `DicountCodes` to `DiscountCode`

## Ticket types
- [x] we need to enforce `available_from` to be earlier than `available_to`

## Tickets
- [x] `discount_code` should be optional
- [x] Does `phone` need to be required?

## Talk
- [x] I think `secondary_speaker` should not be required
- [x] validate that `primary_speaker != secondary_speaker`
- [x] `duration` doesn't work. When I select *30 min* It prints `Select a valid choice. 30 is not one of the available choices.`

## Speaker
- [x] both `url` and `social_url` are required, probably most of the speakers have a website and a social network but all of them? 
- [x] Do speakers have `sponsor` all the time? I'd change it to be optional.

## Receipt
- [x] ~~`title` shouldn't be `name`?~~
- [x] the following field should be IMO optional: `street2`, `companyid`, `taxid`, `vatid`. For example I don't have any of those.
- [x] `amount` decimal field with 2 decimal places should be used

## Schedule
- [x] Shouldn't `description` be `TextField` ?
- [x]  `duration` (in minutes?) currently can be any integer,  range like <0, 300> would make more sense

ricco386:
I just drew DB schema, and drafted models from it so surely it can have some polishing 😄 we haven't implement any logic in models yet. You are right in all of the things, we can implement them as you suggested.

The only thing is that discount_codes should be renamed to discount_code as we use singular for all other models, instead of changing verbose_name_plural.

All url and social_url and sponsor should be optional in speakers model.

We can also change the title from receipt to name, but than we probably should do it also for sponsor, room, location, discount_code(?), event(?), I used title as in most other places I used it...

Regarding schedule description, idea is to use is only for non talks sections (coffee break, etc..) so I dont think there will be a lot of text, but sure lets change it to TextField.

ZoltanOnody:
> We can also change the title from receipt to name, but than we probably should do it also for sponsor, room, location, discount_code(?), event(?), I used title as in most other places I used it...

Ok, I get it now. It just sounded strange so that's why I asked.

[...]

PR #60, opened 2016-09-17
Change models fields #53

ZoltanOnody:
Fix the issues with models in #53. 
I haven't implemented the logic/validation for that issue. It will be in another PR.


ricco386:
I have merged PR, that made some conflict here, please update from upstream and resolve the conflict.

Thanks

Issue #69, opened 2016-09-17
Update sponsor admin class


ricco386:
Create SponsorAdmin class than can do following:

* list view, show fields: title, type, url
* group form fields:
   *  details: all fields can be here...
   *  events: list of events where is sponsoring
   *  soeakers: list of speakers that shopsor had

any other improvements, welcome...

Redpoint1:
working on it

ricco386:
@Redpoint1 thanks

PR #78, opened 2016-09-18
#69 Improved administration for Sponsor

Redpoint1:
FYI i don't like Django's solution for ManyToOne field. Eg. sponsored speakers (Sponsor model). Instead of ManyToMany & ForeignKey field solution (edit in popup, list of already existing speakers) you editing all fields of the related object INLINE. You cant add already existing speakers just new one.


konfera/admin.py
```diff
@@ -70,10 +70,43 @@ class TalkAdmin(admin.ModelAdmin):
...
+        SponsoredSpeakersInline,
+    ]
+
+
+
 admin.site.register(Receipt)
 admin.site.register(Order)
 admin.site.register(Location)
-admin.site.register(Sponsor)
+admin.site.register(Sponsor, SponsorAdmin)
```
ZoltanOnody: Please move the Admin registration just under the `SponsorAdmin` class. Same as `EventAdmin`.

konfera/admin.py
```diff
@@ -70,10 +70,43 @@ class TalkAdmin(admin.ModelAdmin):
...
+    model = Speaker
+    verbose_name = _('Sponsored speaker')
+    verbose_name_plural = _('Sponsored speakers')
+    extra = 1
+
+
+class SponsorAdmin(admin.ModelAdmin):
+    list_display = ('title', 'type', 'url',)
+    list_filter = ('type',)
+    search_fields = ('=title',)
```
ZoltanOnody: What does the `=` means in front of `title` ?

Redpoint1: is a sign for case insensitive searching

ZoltanOnody: thank you :)

konfera/admin.py
Redpoint1: Done

Issue #20, opened 2016-09-10
Call for proposals view


ricco386:
Create form for speaker details and also talk details, that can anyone submit and it will store speaker and talk as a draft in the DB.

TomasHo:
Working on this

TomasHo:
@ricco386 I have worked on this a have a question or two.
I've created 2 ModelForms add_speaker & add_talk, separate views.
Question regarding talk form: Having a form where you can list all registered speakers(through FK relation) doesn't seem the right approach. Any suggestions ?

ricco386:
@TomasHo not entrely sure what you mean with:

> Having a form where you can list all registered speakers(through FK relation) doesn't seem the right approach.

My idea was this:

You crate 2 ModelForms, one for speaker, one for talk, but there will be just one view. New view called cfp, or similar (up to you to pick up the name). In this view you render model forms and in the template you will have one, below other. Once the forms rendered by tempaltes are submitted, you pass the data from request to the ModelForms, both of them has to be valid and if so store data in DB and redirect to some thank you view, or better to landing page and show just [message](https://docs.djangoproject.com/en/1.10/ref/contrib/messages/). However if one of the forms is not valid display the view with error message.

Does this make sense?

SvetlanaM:
for this use case , django formsets is very good, if you need implementation advice

TomasHo:
Thank you both, formsets was the thing I was missing :)

ricco386:
I have merged PR. but in the mean time I have found few isses reported in separate bug #97

PR #87, opened 2016-09-20
#20-cfp-view

TomasHo:


konfera/templates/konfera/cfp_form.html
```diff
@@ -0,0 +1,14 @@
+{% extends 'konfera/base.html' %}
+
+{% block title %}{{ event }}|CFP{% endblock %}
```
Redpoint1: event variable is not defined for this view

konfera/views.py
```diff
@@ -55,3 +58,23 @@ def event_details_view(request, event_slug):
...
+
+def cfp_form_view(request):
+    speaker_form = SpeakerForm(request.POST or None, prefix='speaker')
+    talk_form = TalkForm(request.POST or None, prefix='talk')
+    if speaker_form.is_valid() and talk_form.is_valid():
+        speaker_instance = speaker_form.save()
+        talk_instance = talk_form.save(commit=False)
+        talk_instance.primary_speaker = speaker_instance
+        talk_instance.save()
+        messages.success(request, "Speaker profile successfully created")
```
Redpoint1: it would be nice to display these messages somewhere in base.html

konfera/views.py
```diff
@@ -55,3 +58,23 @@ def event_details_view(request, event_slug):
...
+
+
+def cfp_form_view(request):
+    speaker_form = SpeakerForm(request.POST or None, prefix='speaker')
+    talk_form = TalkForm(request.POST or None, prefix='talk')
+    if speaker_form.is_valid() and talk_form.is_valid():
+        speaker_instance = speaker_form.save()
+        talk_instance = talk_form.save(commit=False)
+        talk_instance.primary_speaker = speaker_instance
+        talk_instance.save()
```
Redpoint1: you dont set "draft" value for the status field

TomasHo:
@Redpoint1 thanks for the review! Should be ok now.

konfera/views.py
```diff
@@ -67,8 +67,9 @@ def cfp_form_view(request):
         speaker_instance = speaker_form.save()
         talk_instance = talk_form.save(commit=False)
         talk_instance.primary_speaker = speaker_instance
+        talk_instance.status = 'cfp'
```
ricco386: hardcoding constants is bad practice. Let`s set the constants free in talks.py (please amend also all other lists).

```python
CFP = 'cfp'

TALK_STATUS = (
    (CFP, 'Call For Proposals'),
)
```
and in this view you can import from CFP and set the constant directly. This way we are always sure we use the same constants, and if it changes in future we change it only at one place.

konfera/forms.py
```diff
@@ -0,0 +1,16 @@
+from django import forms
+from konfera.models import Speaker, Talk
+
+
+class SpeakerForm(forms.ModelForm):
+    class Meta:
+        model = Speaker
+        fields = '__all__'
```
ricco386: Please remove sponsor field, relation between sponsor and speaker will be set by us at this point.

konfera/forms.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class TalkForm(forms.ModelForm):
+
+    class Meta:
+        model = Talk
+        exclude = ['status', 'primary_speaker', 'secondary_speaker']
```
ricco386: exclude event, and update url and view that cfp is part of event and value is predefined from url. This is better UX for user, he has one less field on the form and in case there are multiple evnets user will have the right one.

konfera/views.py
```diff
@@ -55,3 +58,24 @@ def event_details_view(request, event_slug):
     return render(request=request,
                   template_name='konfera/event_details.html',
                   context=context, )
+
+
+def cfp_form_view(request):
+    speaker_form = SpeakerForm(request.POST or None, prefix='speaker')
+    talk_form = TalkForm(request.POST or None, prefix='talk')
```
ricco386: Please pre-define type to talk and duration to 30 minutes. This will improove UX.

ZoltanOnody:
It will be great if you add translation support before merging.

konfera/views.py
ricco386: yes that is good idea, i just mentioned it here as this was highlighted during the review, so i do not forget.

TomasHo:
@ZoltanOnody is that the best practice ? I kind of like to do as much in python syntax as possible. But I can move it to template.

ZoltanOnody:
@TomasHo I think so but let's wait for @ricco386 to be sure.

ricco386:
@TomasHo please use `{% trans %}` in template you have to do `{% load i18n %}` first. I think that is much better practice than store translation strings in view and send into template as variable.

https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#internationalization-in-template-code

TomasHo:
I have some dependancies in #90, so that PR needs to be approved before Travis tests pass.

konfera/forms.py
```diff
@@ -0,0 +1,16 @@
+from django import forms
```
ricco386: You should put a blank line between each group of imports (between django and konfera).

https://www.python.org/dev/peps/pep-0008/#imports

konfera/forms.py
```diff
@@ -0,0 +1,16 @@
...
+from konfera.models import Speaker, Talk
+
+
+class SpeakerForm(forms.ModelForm):
+    class Meta:
```
ricco386: please add blank space so the class definition is consistent with the one below

konfera/views.py
```diff
@@ -1,6 +1,9 @@
+from django.contrib import messages
+from django.http import HttpResponseRedirect
 from django.shortcuts import render, get_object_or_404
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
```
ricco386: You should put a blank line between each group of imports (between django and konfera).

https://www.python.org/dev/peps/pep-0008/#imports

konfera/templates/konfera/cfp_form.html
```diff
@@ -1,11 +1,11 @@
...
 {% block content %}
-    <h2>{{ about_speaker }}</h2>
+    <h2>{% trans "About the speaker" %}</h2>
     <form method="POST" action="">{% csrf_token %}
         {{ speaker_form.as_p }}
-        <h2>{{about_talk }}</h2>
+        <h2>{% trans "About the talk" %}</h2>
         {{ talk_form.as_p }}
-    <input type="submit" value="Save">
+    <input type="submit" value={% trans "Send proposal" %}>
```
ricco386: How about we change the text to:

Submit proposal

TomasHo: Agree, implemented in last commit.

konfera/views.py
```diff
@@ -60,3 +65,25 @@ def event_details_view(request, event_slug):
...
+    speaker_form = SpeakerForm(request.POST or None, prefix='speaker')
+    talk_form = TalkForm(request.POST or None, prefix='talk')
+    if speaker_form.is_valid() and talk_form.is_valid():
+        speaker_instance = speaker_form.save()
+        talk_instance = talk_form.save(commit=False)
+        talk_instance.primary_speaker = speaker_instance
+        talk_instance.status = CFP
+        talk_instance.event = Event.objects.get(slug=event_slug)
+        talk_instance.save()
+        message_text = _("Your talk proposal successfully created")
```
ZoltanOnody: @TomasHo @ricco386 I just realized. Shouldn't be here `ugettext` instead of `ugettext_lazy` ?

Issue #76, opened 2016-09-18
Update order model


ricco386:
In order model we should set purchase date uppon creation so it can have auto_add_now.

Create [save()](https://docs.djangoproject.com/en/1.10/topics/db/models/#overriding-predefined-model-methods) method that will check the status of order and if status changes to paid generate the datetime for payment_date.

ZoltanOnody:
I'll do it

ricco386:
PR merged to master.

Thnak you.

PR #91, opened 2016-09-23
Order

ZoltanOnody:
close issue #76


konfera/models/order.py
```diff
@@ -1,10 +1,14 @@
...
+AWAITING = 'awaiting_payment'
+PAID = 'paid'
+
 ORDER_CHOICES = (
-    ('awaiting_payment', _('Awaiting payment')),
-    ('paid', _('Paid')),
+    (AWAITING, _('Awaiting payment')),
+    (PAID, _('Paid')),
     ('expired', _('Expired')),
     ('cancelled', _('Cancelled')),
```
ricco386: since we have started to use constants in this list, please update all strings into constants, same way you did with AWAITING and PAID.

konfera/tests/test_models.py
```diff
@@ -41,6 +41,16 @@ def test_string_representation(self):
         entry = models.Order(price=155.5, discount=5.5)
         self.assertEqual(str(entry), str(entry.price - entry.discount))
 
+    def test_unpaid_order_save(self):
+        entry = models.Order(price=155.5, discount=5.5)
+        entry.save()
+        self.assertIsNone(entry.payment_date)
```
ricco386: please add assert for default status, so the test yell if someone change it...

Issue #18, opened 2016-09-10
Landing page


ricco386:
On the landing page show list of available events. 
If there is just one event redirect to its details page. 

- [x] #19 Create list events view 
- [x] #17 Show event details view

Redpoint1:
the checkboxes are not checked but tickets are closed ... so the dependencies are solved

ricco386:
@Redpoint1 yes I have updated the issue dependencies as done.

We still need the index page, either display list or details of event...

ZoltanOnody:
I'll do it.

ricco386:
@Redpoint1 @ZoltanOnody thanks PR has been merged to master

PR #93, opened 2016-09-23
Event list redirect

ZoltanOnody:
issue #18 

@ricco386 is this how you mean it?

### DIFF
--- a/konfera/views.py
+++ b/konfera/views.py
@@ -5,7 +5,7 @@ from konfera.models import Event, Talk
 
 
 def event_sponsors_list_view(request, event_slug):
-    event = get_object_or_404(Event, slug=event_slug)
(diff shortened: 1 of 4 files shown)


konfera/views.py
```diff
@@ -31,6 +32,10 @@ def event_speakers_list_view(request, event_slug):
 
 def event_list(request):
     events = Event.objects.all().order_by('date_from')
```
ricco386: I know you haven't write this, but please change this to filter only published events...

ZoltanOnody:
@ricco386 the tests are failing on Django1.10. It looks like `'APP_DIRS': False,` is default setting for django project templates. Any idea how to enforce the app to set `'APP_DIRS': True,` ?

Issue #84, opened 2016-09-19
Create TicketTypeAdmin class


ricco386:
Create TicketTypeAdmin class than can do following:

* filter by: attendee_type
* list view, show fields: title, price, attendee_type, event
* group form fields:
  * details: title, description, price, attendee_type, event
  * availabilty: date_from, date_to

any other improvements, welcome... 

Suggestion, add function that return status of ticket type: not available yet, active, expired depending on current datetime and values defined in model instance. You can add this to admin view. If you create such function, add test for it.

kisso:
I am working on it.

kisso:
Should the status depends only on time?

ricco386:
@kisso yes for now it is sufficient that status is generated by the time compared with from to dates. In the future if we found a use case for different behavior we will refactor.

PR #94, opened 2016-09-24
Ticket type admin

kisso:
issue #84

### DIFF
--- a/konfera/admin.py
+++ b/konfera/admin.py
@@ -130,15 +130,15 @@ admin.site.register(Order, OrderAdmin)
 
 
 class TicketTypeAdmin(admin.ModelAdmin):
-    list_display = ('title', 'price', 'attendee_type', 'event', 'ticket_type_status')
+    list_display = ('title', 'price', 'attendee_type', 'event', 'status')
     list_filter = ('attendee_type',)
-    readonly_fields = ('ticket_type_status',)
+    readonly_fields = ('status',)
     fieldsets = (
         (_('Details'), {
             'fields': ('title', 'description', 'price', 'attendee_type', 'event',)
         }),
         (_('Availability'), {
-            'fields': ('date_from', 'date_to', 'ticket_type_status')
+            'fields': ('date_from', 'date_to', 'status')
         }),
     )
 
--- a/konfera/admin.py
+++ b/konfera/admin.py
@@ -130,14 +130,15 @@ admin.site.register(Order, OrderAdmin)
 
 
 class TicketTypeAdmin(admin.ModelAdmin):
-    list_display = ('title', 'price', 'attendee_type', 'event',)
+    list_display = ('title', 'price', 'attendee_type', 'event', 'ticket_type_status')
     list_filter = ('attendee_type',)
+    readonly_fields = ('ticket_type_status',)
     fieldsets = (
         (_('Details'), {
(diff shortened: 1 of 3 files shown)


konfera/models/ticket_type.py
```diff
@@ -24,6 +25,13 @@ class TicketType(FromToModel):
     def __str__(self):
         return self.title
 
+    def status(self):
```
ricco386: add [@property](https://docs.python.org/3/library/functions.html#property) decorator so we can use status as model property... (see comment below for more info)

konfera/models/ticket_type.py
```diff
@@ -24,6 +25,13 @@ class TicketType(FromToModel):
...
+        if timezone.now() < self.date_from:
+            return _('Not available yet')
+        elif self.date_to < timezone.now():
+            return _('Expired')
+        return _('Active')
```
ricco386: Please add constants. In test you can import constant and compare against it in tests as well, this will work no matter of the language...

```
NOT_AVAILABLE = 'na'
ACTIVE = 'active'
...
```

We might have a dict for resolution:

```
STATUS = {
    NOT_AVAILABLE:  _('Not available yet')
    ACTIVE: _('Active')
    ....
}
```

and function (get_status or something similar) that return translation string, that will be used in admin, or anywhere else where needed...

konfera/tests/test_models.py
```diff
@@ -144,3 +144,21 @@ def test_dates_from_to(self):
         tt.date_to = timezone.now()
         tt.date_from = tt.date_to + datetime.timedelta(+3)
         self.assertRaises(ValidationError, tt.clean)
+
+    def test_unavailable_ticket(self):
+        tt = models.TicketType()
+        tt.date_from = timezone.now() + datetime.timedelta(days=1)
+        tt.date_to = timezone.now() + datetime.timedelta(days=3)
+        self.assertEquals(tt.status(), 'Not available yet')
```
ricco386: Please update test to match the consts...

```
from konfera.models.ticket_type import NOT_AVAILABLE 

self.assertEquals(tt.status, NOT_AVAILABLE)
```

konfera/admin.py
```diff
@@ -129,8 +129,23 @@ class OrderAdmin(admin.ModelAdmin):
...
+class TicketTypeAdmin(admin.ModelAdmin):
+    list_display = ('title', 'price', 'attendee_type', 'event', 'status')
+    list_filter = ('attendee_type',)
+    readonly_fields = ('status',)
+    fieldsets = (
+        (_('Details'), {
+            'fields': ('title', 'description', 'price', 'attendee_type', 'event',)
+        }),
+        (_('Availability'), {
+            'fields': ('date_from', 'date_to', 'status')
```
ricco386: instead of status you can use get_status or something similar defined in comment in model.

ricco386:
@kisso will you have a chance to do the suggested improvements (in near future)? I have done some work on ticket_type model (other sections), so just let me know if you will update the code, or shall i merge your branch into mine and I will do the improvements as part of mine PR. Thanks.

kisso:
@ricco386 I think better way will be that you will merge branch, I am 2nd week at school and we got a lot of tasks to program, so I can't do it right now. Sorry, but when I will be available I will work on project :)

ricco386:
@kisso Thanks.

I have merged the changes into my branch and will be part of different PR.

PR #102, opened 2016-09-28
Room updates

ricco386:
Improoved admin for location and room
Minor model update


konfera/models/room.py
```diff
@@ -3,6 +3,8 @@
 
 class Room(models.Model):
     title = models.CharField(max_length=128)
+    room = models.ForeignKey('Location', related_name='room_locations')
```
ZoltanOnody: `location = ...` not `room`

I'd change the related name to `rooms` it is in my opinion more descriptive when you use it.
```
location = Location.objects.get(...)
location.rooms.all()  # seems better to me
location.room_locations.all()
```

ricco386: :+1: thanks for spotting this. Updated.

ZoltanOnody: I would change the `room =` to `location =`.

ricco386: updated

Issue #98, opened 2016-09-27
Add ability to process FIO bank transfers


ricco386:
FIO bank has API that allows to download transactions statement: http://www.fio.sk/docs/cz/API_Bankovnictvi.pdf

Create a view that can be called by cron every X hours (lets assume 24) and will download transactions and compare them with orders. Each order has to have its own unique numeric ID, this ID is presented to user and he can select bank transfer as a payment method.

We need to go through transactions from the bank and if we find a match of variable symbol and order ID we change the status of the order to PAID.

Talk to me (@RicCo386) about the token for testing.

ZoltanOnody:
I'll do it.

jangondol:
Consider using an external app for this (see [django-fiobank](https://pypi.python.org/pypi/django_fiobank)) or handling payments in a more abstract way. `django-konfera` is getting pretty big and if we start integrating banks *directly*, reuse by others may become difficult.

ricco386:
I agree with @jangondol that if we want to have reusable app this could go into separate app and we shall build hooks in order so we can connect any type of payment provider.

However, I have create this issue as something more difficult for experienced person so he doesnt have to work with simple tasks, and he can leave the simple tasks to other less experienced. So in this case I dont mind if @ZoltanOnody will create his directory with urls, views, etc. and do the logic there: at the moment one view that will be called periodically. Later we can re-factor and move it to separate app if necessary (this task is quite ahead of the development at this moment). If more people should work on this we can not separate everything at this point as it will be pain to get them all on board.

OK so lets do it properly (at least easy to refactor properly) and I am thinking about signals: 
 * Create directory payments, inside django-konfera
 * Create signals.py where we will collect all signal handlers
 * Create new view: /payments/check that will emit signal, in this signal we can have query of all orders that shall be checked (we need to have a thought about this bit)
 * Create fio.py what will receive signal and process all orders

If we do it like this anyone can write his own signal receiver and connect any type of processing, at the same time we can easily refactor fio (or whole payments) into separate app. This is just an initial suggestion, please feel free to come up with better one. I need to sleep about this and have a thought if this is correct (for now it is just info for @ZoltanOnody what way we might go with this).

ricco386:
PR merged, with sync functionality. We might improove in future, but separate issue will be created for it.

PR #103, opened 2016-09-28
Add FioBank payments checker

ZoltanOnody:
Issue #98 

TODO:
- [x] Add `fiobank` to travis so test won't fail
- [x] [Fix test ](https://github.com/pyconsk/django-konfera/compare/master...ZoltanOnody:fio_payments?expand=1#diff-b945724f5d95ac410f30696347136477R92)
- [x] [Froze time for test](https://github.com/pyconsk/django-konfera/compare/master...ZoltanOnody:fio_payments?expand=1#diff-b945724f5d95ac410f30696347136477R93)
- [x] I am not sure how payments in different currencies will wok. Should we [check whether currency is EUR](https://github.com/pyconsk/django-konfera/compare/master...ZoltanOnody:fio_payments?expand=1#diff-5d7894b454aeca4c3fcc7318f3103f52R16) ?

====

NEW TODO:
- [x] use `.variable_symbol` instead of  `.pk` 
- [x] add logs when payment is processed
- [x] ~~send mail to attendee when his payment is processed~~ I'll create a separate issue for this later
- [x] ~~select only the payments for last 3/4 days for [processed payments](https://github.com/pyconsk/django-konfera/pull/103/files#diff-5d7894b454aeca4c3fcc7318f3103f52R33) - now we collect only the last 3 days of payments from FioBanka so getting all the processed payments is unnecessary~~ (premature optimalization)


ricco386:
Currently our model does not support different currencies.

Eg. everything has just an amount so in our system doesn't matter what currency is used as long as you use only one currency through out the whole system. Eg. if you have FIO in CZK and all ticket prices, orders, everything will be in CZK should be OK. For our situation we use and accept only EUR, and will use all sums in EUR (we do have CZ account but if someone send CZ krown, bank will automatically convert to EUR and add EUR to our account, so this will be ok even in your case, we might end up to have bit more, or less depending on exchange rate, but in the past we didnt care about few cents difference).

I think we can support multiple currencies in future, but for that we will have to update our internal models.

ZoltanOnody:
Thank you. Can it stay as it is? So if the bank converts the money to EUR and we get less we will change the order's status manually or should I change it so we accept everything in a  (for example) 1% error rate?

ricco386:
I like the idea with the error rate, please do it but in a way that we can easily change the percentage, eg. setting would be nice.

ZoltanOnody:
It's basically done. We will need to invoke the `payments.utils.check_payments_status` function regularly via CRON. It checks & changes the status of orders if it can associate it with a payment from FioBank payments list. 

@ricco386 please review it.

payments/utils.py
```diff
@@ -0,0 +1,47 @@
...
+from fiobank import FioBank
+
+from konfera.models import Order
+from konfera.models.order import AWAITING, PAID
+
+
+DATE_FORMAT = '%Y-%m-%d'
+
+
+def _is_order_paid(order, payments):
```
ricco386: Payments are important and in this case i think we shall log everything. Please consider adding log:

* payment was found but it is outside of error rate (WARNING)
* payment status has changed (INFO)

There might be monitoring in place which will alerts warnings.

example/example/settings.py
```diff
@@ -123,3 +125,6 @@
 
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
 MEDIA_URL = '/media/'
+
+FIO_BANK_TOKEN = 'token'
+PAYMENT_ERROR_RATE = Decimal('0.0')
```
ricco386: consider integer field: 0 to 100 as that is also used in DB for discount (so we are more consistent, and is more user friendly to setup). Originally I thought we can have 3.5% error rate...

ZoltanOnody:
@ricco386 the request changes were made. Except the *send an email to the user* part. I'll create a separate issue for that as soon as this PR will be merged. 

Please review this PR.

payments/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    - add payment to ProcessedTransaction
+    """
+    amount_to_pay = order.left_to_pay - payment['amount']
+
+    if amount_to_pay <= order.to_pay * Decimal(settings.PAYMENT_ERROR_RATE / 100):
+        order.status = PAID
+
+        msg = 'Order(id={order_id}) was paid in payment with transaction_id={transaction_id}'.format(
+            order_id=order.pk, transaction_id=payment['transaction_id'])
+        logger.warning(msg)
```
ricco386: Order was paid and everything is alright I think log level info level is sufficient here...

payments/utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+DATE_FORMAT = '%Y-%m-%d'
+
+
+logger = logging.getLogger(__name__)
+
+
+def _get_last_payments():
+    """ Get list of payments for last three days from FioBank """
+    client = FioBank(token=settings.FIO_BANK_TOKEN)
```
ricco386: this is third party connector (either here or in period method), I am just curious what happen if lets say Fio is down, or our server is running but offline...

I just had a thought what if we modify it to be executable via button (lets say we have button in admin area, process FIO payments) and if FIO is down for maintance for example, we want to show nice message instead of exception... How about we wrap it in try catch block and log all exceptions, this way we will have data in case problem in FioBank occurs, for the moment we can have exception at least we will test the third party app a little before  real usage...

payments/models.py
```diff
@@ -0,0 +1,6 @@
+from django.db import models
+
+
+class ProcessedTransaction(models.Model):
```
ricco386: Just thinking if we do store transaction, do we want to store more metadata? Such as bank account from which we received payment, amount received, other useful info that can be used on invoice?

payments/models.py
```diff
@@ -0,0 +1,6 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
```
ricco386: Since this app is already depended on konfera why dont we use KonferaModel, as it already have date_created field...?

ricco386:
:+1: for large test set...
There are things to discuss, and also minor updates... @ZoltanOnody please have a look.

ricco386:
@ZoltanOnody PR merged, thanks

Issue #75, opened 2016-09-18
Update ticket model save logic


ricco386:
Once ticket is created and stored in DB order should be automatically generated if it is not part of the ticket.

Create [save()](https://docs.djangoproject.com/en/1.10/topics/db/models/#overriding-predefined-model-methods) method that will check if order is not set and if so, create new order. Set status to awaiting_payment, price is copied from ticket type, and discount is calculated from percentage if discount it set.

kapucko:
I'm taking also this one.

SvetlanaM:
pre_save and post_save signals are also good https://docs.djangoproject.com/en/1.10/topics/signals/ on this

ricco386:
PR merged

PR #104, opened 2016-09-29
#75 ticket save logic

kapucko:
model logic updated + 1 test


konfera/models/ticket.py
```diff
@@ -2,6 +2,8 @@
 from django.utils.translation import ugettext_lazy as _
 
 from konfera.models.speaker import TITLE_UNSET, TITLE_CHOICES
+from konfera.models.order import Order
+from django.utils import timezone
```
ricco386: please move django import together, and leave one blank line between django and konfera imports, so it is according PEP8.

https://www.python.org/dev/peps/pep-0008/#imports

konfera/models/ticket.py
```diff
@@ -30,3 +32,14 @@ def __str__(self):
             first_name=self.first_name,
             last_name=self.last_name
         ).strip()
+
+    def save(self, *args, **kwargs):
+        if not hasattr(self, 'order'):
+            discount = 0
+            if self.discount_code:
```
ricco386: add blank line before if for better readability.

konfera/models/ticket.py
```diff
@@ -30,3 +32,14 @@ def __str__(self):
...
+                discount = self.type.price * self.discount_code.discount/100
```
ricco386: Please create function discount_calculator() with logic above to calculate and return discount. We will most likely update this logic a bit more in near future.

konfera/models/ticket.py
```diff
@@ -30,3 +32,14 @@ def __str__(self):
...
+            order = Order(price=self.type.price, discount=discount, status='awaiting_payment',
+                          purchase_date=timezone.now())
+            order.save()
+            self.order = order
+        super(Ticket, self).save(*args, **kwargs) # Call the "real" save() method.
```
ricco386: Comments are good, but in this case it is obvious what you are doing, therefore I think not necessary here. Besides you are missing one space there, so it breaks PEP8 and travis continuous deployment. Either fix PEP8 or remove comment.

konfera/models/ticket.py
```diff
@@ -30,3 +32,14 @@ def __str__(self):
             first_name=self.first_name,
             last_name=self.last_name
         ).strip()
+
+    def save(self, *args, **kwargs):
+        if not hasattr(self, 'order'):
+            discount = 0
+            if self.discount_code:
+                discount = self.type.price * self.discount_code.discount/100
+            order = Order(price=self.type.price, discount=discount, status='awaiting_payment',
```
ricco386: Using string instead of constant is bad practice. Please import from order AWAITING and use constant for status.

konfera/tests/test_models.py
```diff
@@ -133,6 +133,28 @@ def test_string_representation(self):
             '%s %s %s' % (dict(TITLE_CHOICES)[entry.title], entry.first_name, entry.last_name)
         )
 
+    def test_automatic_order_generator(self):
+        location = models.Location(title='test_title', street='test_street', city='test_city', postcode=000000,
+                                   state='test_state', capacity=20)
+        location.save()
+        event = models.Event(title='test_event', description='test', event_type='meetup', status='published',
```
ricco386: Please consider importing status and use const. In test I dont mind that much, but it is good practice to use const as you are preventing typos, and if in future we need to change it, we'll do it at one place.

konfera/tests/test_models.py
```diff
@@ -133,6 +133,28 @@ def test_string_representation(self):
...
+                             location=location, date_from=timezone.now(), date_to=timezone.now())
```
ricco386: you can store timezone.now() in variable and use it everywhere. Reason for this is that you pre-cache the value so you don't need to ask for it multiple times, and other thing is that the values would be different as milliseconds would change while this will be processed.

kapucko:
Thanks for feedback :) should be fixed now...

konfera/models/ticket.py
```diff
@@ -33,13 +33,15 @@ def __str__(self):
             last_name=self.last_name
         ).strip()
 
+    def discount_calculator(self):
+        if self.discount_code:
+            return self.type.price * self.discount_code.discount / 100
+        return 0
```
ricco386: If you can avoid have multiple return in function, please do that in future. It will improve readability and reduce code complexity. You can do it like this:

```
def discount_calculator(self):
       discount = 0

       if self.discount_code:
           discount = self.type.price * self.discount_code.discount / 100
       return discount
```

konfera/tests/test_models.py
```diff
@@ -134,22 +134,22 @@ def test_string_representation(self):
...
 
     def test_automatic_order_generator(self):
+        time = timezone.now()
         location = models.Location(title='test_title', street='test_street', city='test_city', postcode=000000,
                                    state='test_state', capacity=20)
         location.save()
-        event = models.Event(title='test_event', description='test', event_type='meetup', status='published',
-                             location=location, date_from=timezone.now(), date_to=timezone.now())
+        event = models.Event(title='test_event', description='test', event_type='meetup',
+                             status=models.event.PUBLISHED, location=location, date_from=time, date_to=time)
```
ricco386: Probably better to import at the top of the file and use only constant here, also I noticed event_type above can used const as well...

Issue #83, opened 2016-09-19
Create DiscountCodeAdmin class


ricco386:
Create DiscountCodeAdmin class than can do following:

* list view, show fields: title, discount, ticket_type, usage
* group form fields:
  * details: title, usage, hash
  * discount: discount, usage
  * availabilty: date_from, date_to

any other improvements, welcome... Suggestion, add function that return status of discount: not available yet, active, expired depending on current datetime and values defined in model instance. You can add this to admin view. If you create such function, add test for it...

ZoltanOnody:
done in #106

PR #106, opened 2016-09-30
From to model updates

ricco386:
I have done som work on abstract model which resulted in larger commit. There is work from multiple issues (sorry), but it was logically related.

Merge wrok from PR #95 issues #83 #84 improoved tests and added some sample data...


konfera/models/ticket_type.py
```diff
@@ -16,14 +28,47 @@
...
     event = models.ForeignKey('Event')
 
     def __str__(self):
         return self.title
 
+    def _get_current_status(self):
+        now = timezone.now()
+        status = ACTIVE
+
+        if not self.date_from or not self.date_to or now < self.date_from:
```
ZoltanOnody: Is this check `not self.date_from or not self.date_to` necessary? I think every TicketType will have `date_from` because either the user sets it or the `clean` does. The same applies for `date_to`.

ricco386: Nope. Currently we have status  in admin, and if admin is adding new ticket type we got an exception...

`TypeError at /admin/konfera/tickettype/add/

unorderable types: datetime.datetime() < NoneType()`

So either we need to hide it in admin or leave the condition as it is...

PR #115, opened 2016-10-02
Base template

ricco386:
This is just a first step to copy the design we had in PyCon SK 2016 as a base template.


konfera/static/konfera/css/pycon.css
```diff
@@ -0,0 +1,656 @@
...
+
+hr {
+  border-color:#ececec;
+}
+input, select {
+  width: 98%;
+  margin: 1px 0;
+}
+button {
+ outline: 0;
```
ZoltanOnody: In most of the file we use 2 spaces, here is just one (same applies for some other lines). Let's keep it consistent.

konfera/utils.py
```diff
@@ -0,0 +1,13 @@
+from django.conf import settings
+
+
+def collect_view_data():
```
ZoltanOnody: Instead of creating a util and invoking it in every single view. I'd create a [context processor](https://docs.djangoproject.com/en/1.10/ref/templates/api/#writing-your-own-context-processors)

konfera/templates/konfera/base.html
```diff
@@ -11,17 +13,77 @@
...
+            <li><a href="mailto:info%40pycon.sk"><span class="glyphicon glyphicon-envelope" aria-hidden="true"></span> info@pycon.sk</a></li>
+            <li class="hide"><a href="../sitemap.xml">sitemap.xml</a></li>
+            -->
+          </ul>
+
+        </div>
+      </div>
+      <div class="row">
+        <div class="col-md-12">
+          <small class="text-center"><a href="https://www.python.org/psf/trademarks/pycon/" target="_blank">PyCon™</a>is a trademark for worldwide conference activities claimed by the <a href="https://www.python.org/psf-landing/" target="_blank">Python Software foundation</a>.</small>
```
ZoltanOnody: Let's keep the line length bellow 120 chars in templates as well. This is hard to read.

ricco386:
all done

PR #117, opened 2016-10-03
Added uuid for certain fields, updated migration

ricco386:
Added uuid for certain fields, updated migration
Minor admin updates


konfera/models/ticket_type.py
```diff
@@ -33,12 +35,25 @@
     (AID, _('Aid')),
 )
 
+PUBLIC = 'public'
+PRIATE = 'private'
```
ZoltanOnody: typo in a variable name

ricco386:
alright, lets use it instead of primary key than.

konfera/models/order.py
```diff
@@ -1,10 +1,13 @@
+import uuid
```
ZoltanOnody: You don't need to import `uuid` in models anymore (only in `abstract.py`).

Issue #97, opened 2016-09-26
CfP issues


ricco386:
@TomasHo I have found few issues with CfP view while testing, please have a look and fix.

* if event does not exists return 404 instead of form, as the store results in error anyway.
* once forms are saved, redirect to event details (not list of events).
* change url from add_cfp just to cfp

ricco386:
You can pull from my branch [97-CfP_issues](https://github.com/ricco386/django-konfera/tree/97-CfP_issues) I have created test for both of this issues.

ricco386:
Unfortunately I have removed the branch so you can copy the test from commit 682b108.

I have changed the template to use tables as form, but I have decided that as_p is more flexible so please re-factor view to use p instead of table.

ricco386:
I will work on this

ricco386:
PR merged

PR #140, opened 2016-10-05
97 cfp issues

ricco386:
#97 cfp issues


ricco386:
Finally it works :smile: I found very strange travis behaviour... Originally I had an logical issue in the test, so they were failing... In the mean time another PR modified urls, but once I have modified code in the branch tests were comparing to newly modified urls, so it seems as travis is running tests on upstream master.

pull request has to have allways up to date master, even if there was another merge that does not change anything in PR (so auto merge would work).

PR #142, opened 2016-10-06
Schedule

### LINKED ISSUE #129: Create schedule view
Create schedule view, if event has defined some schedule, create view that will display it. Please add also design similar to the one at https://2016.pycon.sk/sk/schedule-friday.html

ZoltanOnody:
close #129 

image


ZoltanOnody:
I fixed the pep8 problem. Only 2 tests are failing which you've already fixed. So I thing it can be merged, right?

PR #153, opened 2016-10-08
updating readme to reflect pyvenv instead of virtualenv

bambielli:
Since the project only supports python3, and python3 comes with its own version of virtualenv in stdlib, we should just point to that instead of instructing new developers to install virtualenv from pip.

TIL!


ZoltanOnody:
It makes sense to me but let's wait for @ricco386 :)

ricco386:
@bambielli thanks

PR #154, opened 2016-10-08
Validation logic when saving a Ticket

### LINKED ISSUE #148: Ticket save validation
Creating ticket allows to store not valid data. If you have multiple ticket types for multiple events and discount codes in multiple events. System does not check if you do not mix discount codes and different events. Update save method to check if discount code belongs to ticket type

bambielli:
fixes #148.
This pull request adds validation logic when saving a ticket, either by submitting a form or by saving manually, to ensure that the selected discount_code for the ticket is applicable to the ticket_type of the ticket being created.

I added a test in test_models.py to confirm that the logic works as intended.

Let me know if there are any questions!

### DIFF
--- a/konfera/models/ticket.py
+++ b/konfera/models/ticket.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 from django.utils import timezone
 
@@ -44,7 +45,14 @@ class Ticket(KonferaModel):
             return self.type.price * self.discount_code.discount / 100
         return 0
 
+    def clean(self):
+        if self.discount_code and self.discount_code.ticket_type != self.type:
+            raise ValidationError(
+                {'discount_code': _('Discount code is not applicable for the selected ticket type.')})
+        super().clean()
+
     def save(self, *args, **kwargs):
+        self.clean()
         if not hasattr(self, 'order'):
             discount = self.discount_calculator()
             order = Order(price=self.type.price, discount=discount, status=AWAITING, purchase_date=timezone.now())
--- a/konfera/tests/test_models.py
+++ b/konfera/tests/test_models.py
@@ -178,6 +178,27 @@ class TalkTest(TestCase):
 
 
 class TicketTest(TestCase):
+    def setUp(self):
+        time = timezone.now()
+        location = models.Location(title='test_title', street='test_street', city='test_city', postcode='000000',
+                                   state='test_state', capacity=20)
+        location.save()
+        event = models.Event(title='test_event', description='test', event_type='meetup',
+                             status=models.event.PUBLISHED, location=location, date_from=time, date_to=time)
+        event.save()
+        self.ticket_type = models.TicketType(title='test', description='test', price=100, event=event,
+                                             date_from=time, date_to=time)
+        self.ticket_type.save()
+        self.ticket_type_2 = models.TicketType(title='secondType', description='otherType', price=7, event=event,
+                                               date_from=time, date_to=time)
+        self.ticket_type_2.save()
+        self.discount_code = models.DiscountCode(title='test discount', hash='test', discount=60,
+                                                 date_from=time, date_to=time, usage=1,
+                                                 ticket_type=self.ticket_type)
+        self.discount_code.save()
+        self.discount_code_2 = models.DiscountCode(title='discount_2', hash='otherCode', discount=4,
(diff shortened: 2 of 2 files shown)


konfera/models/ticket.py
```diff
@@ -44,7 +45,14 @@ def discount_calculator(self):
...
         return 0
 
+    def clean(self):
+        if self.discount_code and self.discount_code.ticket_type != self.type:
+            raise ValidationError(
+                {'discount_code': _('Discount code is not applicable for the selected ticket type')})
+        super(Ticket, self).clean()
+
     def save(self, *args, **kwargs):
+        self.clean()
```
bambielli: This makes it so the clean logic is run when someone tries to save a ticket object programmatically.

konfera/models/ticket.py
```diff
@@ -44,7 +45,14 @@ def discount_calculator(self):
             return self.type.price * self.discount_code.discount / 100
         return 0
 
+    def clean(self):
+        if self.discount_code and self.discount_code.ticket_type != self.type:
+            raise ValidationError(
+                {'discount_code': _('Discount code is not applicable for the selected ticket type')})
```
ZoltanOnody: To keep the error messages consistent, please add a period to the end.

konfera/models/ticket.py
```diff
@@ -44,7 +45,14 @@ def discount_calculator(self):
...
+        super(Ticket, self).clean()
```
ZoltanOnody: We support only Python3 so `super().clean()` is enough.

konfera/tests/test_models.py
```diff
@@ -193,27 +214,38 @@ def test_string_representation(self):
...
-        self.assertEquals(ticket.order.price, ticket_type.price)
+        self.assertEquals(ticket.order.price, self.ticket_type.price)
         self.assertEquals(ticket.order.discount, 60)
 
+    def test_clean_and_save_ticket_discount_code(self):
+
+        # if cleaned and saved with no discount_code, they should both be successful
+        ticket_no_code = models.Ticket(status='requested', title='mr', first_name="test", last_name="Test",
+                                       type=self.ticket_type, email='test@test.com', phone='0912345678')
+        ticket_no_code.clean()
```
ZoltanOnody: You put the `.clean()` to `.save()` so invoking it here is not necessary. Same applies for line 238.

konfera/tests/test_models.py
```diff
@@ -178,6 +178,27 @@ def test_string_representation(self):
 
 
 class TicketTest(TestCase):
+    def setUp(self):
+        time = timezone.now()
+        location = models.Location(title='test_title', street='test_street', city='test_city', postcode=000000,
```
ZoltanOnody: I know this *bug* was here before but `postcode` is a `CharField` so `'00000'` would be better.

konfera/tests/test_models.py
```diff
@@ -178,6 +178,27 @@ def test_string_representation(self):
...
+                                   state='test_state', capacity=20)
+        location.save()
+        event = models.Event(title='test_event', description='test', event_type='meetup',
+                             status=models.event.PUBLISHED, location=location, date_from=time, date_to=time)
+        event.save()
+        self.ticket_type = models.TicketType(title='test', description='test', price=100, event=event,
+                                             date_from=timezone.now(), date_to=timezone.now())
```
ZoltanOnody: Just curious. Is there a reason why we use `timezone.now()` here instead of the variable `time` ?

konfera/models/ticket.py
bambielli: ok

konfera/models/ticket.py
bambielli: Hmm I didn't realize that you didn't need to pass the class + context in python3! TIL. I'll make the change.

konfera/tests/test_models.py
bambielli: Not sure, I just copied and pasted from a different test that was constructing a new object :P I think we could use the variable. I'll change it here.

konfera/tests/test_models.py
bambielli: I took out the calls to "clean" throughout the test. You're right, since clean is being called from save, we can just test it indirectly by calling save.

bambielli:
@ZoltanOnody I made the changes you suggested. Please let me know if there is anything else you notice! 

Can you also tag this pull request with the "hacktoberfest" tag?

ZoltanOnody:
Looks good to me. Thank you.

PR #162, opened 2016-10-11
change models ( with choices)

PKpacheco:


konfera/models/event.py
```diff
@@ -4,33 +4,23 @@
...
-)
-
-
 class EventManager(models.Manager):
     def published(self):
-        return self.get_queryset().filter(status=PUBLISHED)
+        return self.get_queryset().filter(status='published')
 
 
 class Event(FromToModel):
```
ricco386: add also constants inside the class:

```
CONFERENCE = 'conference'
MEETUP = 'meetup'

DRAFT = 'draft'
...
```

reason for this is that everywhere in the code we would refer to constant. We do it because you will not have a typo in string (it is easy to make such mistake), and second if we need to change the value of the constant we do it and one place and everywhere in the code it is changed. 

For example if we change 'draft' to 'drafted' you would have to update all filters, validation methods, everywhere you compared to string, but if you use cont you change it just here.

konfera/models/event.py
```diff
@@ -4,33 +4,23 @@
...
+    EVENT_TYPE_CHOICES = (
+        ('conference', _('Conference')),
```
ricco386: please do not use string but constant defined above:

```
(CONFERENCE, _('Conference')),
```

konfera/models/event.py
```diff
@@ -4,33 +4,23 @@
...
-    (PUBLISHED, _('Published')),
-    (PRIVATE, _('Private')),
-    (EXPIRED, _('Expired')),
-)
-
-
 class EventManager(models.Manager):
     def published(self):
-        return self.get_queryset().filter(status=PUBLISHED)
+        return self.get_queryset().filter(status='published')
```
ricco386: Please do not compare to string, use constant instead (reason explained in Event model).

```
Event.PUBLISHED
```

konfera/models/core.py
```diff
@@ -0,0 +1,79 @@
+EVENT_TYPE_CHOICES = (
+    ('conference', ('Conference')),
```
ricco386: you probably forgot _ and import at the begging of the file as 'Conference' was meant to be translation string.

```
from django.utils.translation import ugettext_lazy as _

('conference', _('Conference')),
```

ricco386:
@PKpacheco how it is going with issue solving? Do you need any help? I have merged few other PRs in the mean time, so please update your fork.

ricco386:
@PKpacheco just wondering if you will have time ti finish this? Or if I should ask someone else to do it?

ZoltanOnody:
Closing due to lack of follow up.

Issue #120, opened 2016-10-03
volunteers registration (view and form) updates


ricco386:
view updates:
 * after storing data in DB  show successful message
 * make sure ticket_type is public

form updates:
 * Make description field required.
 * Add help text for description: Please tell us something about yourself.

ricco386:
Once you are at it please also do this improovments:

In view, please get rid of the if request.method == "POST" line with form = VolunteerRegistrationForm(request.POST or None) this way we can get rid of one if statement so the code is easier to read (more if statement worse Cyclomatic complexity score results). See cfp_form_view as an example.

In template update the button to be nicer and uses Bootstrap styling. See egistration_form.html as an example.

ricco386:
There is an issue #125 that changes the directory structure to avoid conflicts, lets do the parent issue first.

andela-ijubril:
Hi @ricco386, I would like to takea shot at this.

ricco386:
Hello @andela-ijubril sure go ahead. If you need more info or something is not clear just ask. Thanks

andela-ijubril:
Hi @ricco386, I need more info on this issue. Thanks

ricco386:
PR #130 has been closed thanks @andela-ijubril

PR #164, opened 2016-10-12
solved Issue #120

andela-ijubril:


konfera/register/views.py
```diff
@@ -21,6 +21,10 @@ def register_volunteer(request, slug):
             new_ticket.type = volunteer_ticket_type
             new_ticket.save()
 
+            if volunteer_ticket_type.accessibility != 'public':
```
ricco386: It seems, I havent explained the issue very well. What I meant was that the initial check in this file (get_object_or_404 should have one more search attribute accessibility). Eg. current setup would also allow volunteer registration event if there  is only private ticket, you update so it will only allow if there is public ticket.

Your change would modify existing ticket from private to public... Please revert back, other thing I noteiced is using string for constant comparation is bad practice in konfera.models.ticket_type there is PUBLIC constant which should be used in such case.

konfera/templates/konfera/event_details.html
```diff
@@ -8,6 +8,12 @@
 
 
 {% block content %}
+
+    <div class="alert alert-success" id="success-alert">
```
ricco386: we already have this in base.html no need for repeating code, please remove.

konfera/register/views.py
```diff
@@ -21,6 +21,10 @@ def register_volunteer(request, slug):
...
+                volunteer_ticket_type.accessibility = 'public'
+                volunteer_ticket_type.save()
+
```
ricco386: Add message here. Have a look in events/views.py in cfp_form_view how we add messages.

PR #165, opened 2016-10-12
Issue 157: Ticket discount code usage

mimmon:
Added method to lower (subtract) the usage of discount code.
After the order is saved, the usage is subtracted.
Not sure, but one of the validations may be redundant.


konfera/models/discount_code.py
```diff
@@ -23,6 +23,13 @@ class DiscountCode(FromToModel):
     def __str__(self):
         return self.title
 
+    def sub_usage(self, num=1):
```
ricco386: change to get_usage and upfate functionality to filter all usages in tickets.

konfera/models/ticket.py
```diff
@@ -40,15 +40,18 @@ def __str__(self):
...
 
     def clean(self):
-        if self.discount_code and self.discount_code.ticket_type != self.type:
-            raise ValidationError(
-                {'discount_code': _('Discount code is not applicable for the selected ticket type.')})
+        if self.discount_code:
+            if self.discount_code.ticket_type != self.type:
+                raise ValidationError(
+                    {'discount_code': _('Discount code is not applicable for the selected ticket type.')})
+            if self.discount_code.usage <= 0:
```
ricco386: compare to get_usage function

konfera/models/ticket.py
```diff
@@ -58,4 +61,6 @@ def save(self, *args, **kwargs):
             order = Order(price=self.type.price, discount=discount, status=AWAITING, purchase_date=timezone.now())
             order.save()
             self.order = order
+            if discount:
```
ricco386: we dont need this after we create get_usage function

ricco386:
Hey @mimmon will you be able to do the updates I have described? Please let me know either way. Thanks

mimmon:
Sorry, been busy lately. I planned it for yesterday, but did not manage. I
would like to do it, however, so I will start today. Will let you know in
case I got stuck.

ricco386:
Sure, take your time. If you need anything let's discuss here or at gitter. 

Thanks

konfera/models/discount_code.py
```diff
@@ -23,6 +24,14 @@ class DiscountCode(FromToModel):
     def __str__(self):
         return self.title
 
+    def get_usage(self):
+        tickets = Ticket.objects.filter(discount_code__hash=self.hash)
```
ricco386: please add blank line below tickets, for better readability

konfera/models/discount_code.py
```diff
@@ -23,6 +24,14 @@ class DiscountCode(FromToModel):
     def __str__(self):
         return self.title
 
+    def get_usage(self):
```
ricco386: How about we rename to something more obvious, like: `issued_tickets`
Also we can mar this function with `@property` decorator, this way it would be possible to call it `DiscountCode.issued_tickets`

konfera/models/ticket.py
```diff
@@ -53,6 +52,8 @@ def clean(self):
             exist_ticket = self.order.ticket_set.first()
             if exist_ticket and exist_ticket.type.event.id != self.type.event.id:
                 raise ValidationError(_('All tickets must be for the same event.'))
+        if self.discount_code and self.discount_code.is_available is False:
```
ricco386: This is valid code, but I think if you change it to:
`if self.discount_code and not self.discount_code.is_available:`
it would be more 'pythonic'...

ricco386:
LGTM @mimmon thank you

PR #166, opened 2016-10-15
Updated Event.slug field to be unique and 128 chars long

dn0:
Updated Event.slug field to be unique and have the same length as Event.title field.
This fixes possible bugs with long titles. Also the slug must be unique because it is used a event identifier.


ricco386:
@dn0 thanks, merged into master

PR #167, opened 2016-10-18
Add a view for private registration

### LINKED ISSUE #122: private registration view
Add a view that will check uuid in the url, based on this uuid form for ticket is rendered.
 
* Make sure TicketType UUID get from url and ACCESSIBILITY is Private
* TicketType has to be in available date range
* ticket with price is in requested state, everyone else in state registered
* Those in requested state have description required field in the form

simonsmiley:
Fixes #122


simonsmiley:
I guess the views could be merged together (possibly class based views?) But this should do the job...

simonsmiley:
should I add tests?

konfera/register/forms.py
```diff
@@ -3,13 +3,13 @@
 from konfera.models import Ticket
 
 
-class VolunteerRegistrationForm(forms.ModelForm):
+class RegistrationForm(forms.ModelForm):
     class Meta:
         model = Ticket
         fields = ('title', 'first_name', 'last_name', 'email', 'phone', 'description')
 
     def __init__(self, *args, **kwargs):
```
ricco386: Add a parameter that would decide if description field should be required of not. By default we can have description optional field.

konfera/register/views.py
```diff
@@ -4,8 +4,37 @@
...
+
+def private_registration(request, event_slug, ticket_uuid):
+    context = dict()
+    event = get_object_or_404(Event, slug=event_slug)
+    ticket_type = get_object_or_404(TicketType, event=event.id, uuid=ticket_uuid, accessibility=PRIVATE)
+    if ticket_type.status != ACTIVE:
+        messages.error(request, "This ticket type is not available")
+        return redirect('event_details', event_slug)
+
+    form = RegistrationForm(request.POST or None)
```
ricco386: If related ticket_type is in (VOLUNTEER, PRESS, AID) make description field required, for other ticket types it can be optional.

ricco386:
@simonsmiley for the tests, we do encourage people to write a tests for the changes they have done, but it is not mandatory...

konfera/register/views.py
```diff
@@ -4,8 +4,37 @@
...
+from konfera.models.ticket_type import TicketType, VOLUNTEER, PUBLIC, PRIVATE, ACTIVE
+from konfera.register.forms import RegistrationForm
+
+
+def private_registration(request, event_slug, ticket_uuid):
+    context = dict()
+    event = get_object_or_404(Event, slug=event_slug)
+    ticket_type = get_object_or_404(TicketType, event=event.id, uuid=ticket_uuid, accessibility=PRIVATE)
+    if ticket_type.status != ACTIVE:
+        messages.error(request, "This ticket type is not available")
```
simonsmiley: The strings in the class where not translated, so I assumed it was just english.

Added the translations...

simonsmiley:
I updated the pr to match what you requested

ricco386:
@simonsmiley PR merged, thanks.

PR #169, opened 2016-10-20
Fix error when comparing status of ticket types

simonsmiley:
status returns a translated string, so it can't be used to check the status.


ricco386:
@simonsmiley thanks for spotting this. PR merged.

PR #170, opened 2016-10-20
Calculate price and discount on orders automatically

### LINKED ISSUE #149: Update Order save function
Currently if we create order via admin we can and add multiple tickets. Update order save method to re-calculate price and discount for all tickets together.

simonsmiley:
Fixes #149 
I decided to put the code in the ticket save method because if tickets are updated in some other way the order will reflect the change.


konfera/models/ticket.py
```diff
@@ -58,4 +58,12 @@ def save(self, *args, **kwargs):
             order = Order(price=self.type.price, discount=discount, status=AWAITING, purchase_date=timezone.now())
             order.save()
             self.order = order
-        super(Ticket, self).save(*args, **kwargs)
+            super(Ticket, self).save(*args, **kwargs)
+        else:
+            super(Ticket, self).save(*args, **kwargs)
+            self.order.price = 0
+            self.order.discount = 0
+            for ticket in self.order.ticket_set.all():
```
ricco386: we can have function in order `recalculate_tickets` that will contain this functionality.

konfera/models/ticket.py
```diff
@@ -58,4 +58,12 @@ def save(self, *args, **kwargs):
             order = Order(price=self.type.price, discount=discount, status=AWAITING, purchase_date=timezone.now())
             order.save()
             self.order = order
-        super(Ticket, self).save(*args, **kwargs)
+            super(Ticket, self).save(*args, **kwargs)
+        else:
+            super(Ticket, self).save(*args, **kwargs)
```
ricco386: Please update the code so it calls super just once at the end of this function.

In else block we can just call `order.recalculate_tickets()`

simonsmiley: I wanted to do that, but decided against it.
In case an order has to be created first, there will only be one ticket, so that the price and discount are the values of the one ticket. 
If however an order exists, we don't actually know if the current ticket is new or is being edited. We would have to loop over the tickets and exclude the current ticket if it exists. 
By saving the ticket first, we can be sure that the we calculate the correct price. If you want I can add a comment about that.

ricco386: I see your point.

The original issue #149 was suggesting that login would be in order. I think it is better solution to have `recalculate_ticket_price` (or similar name) function in order. This function would the logic you have below.

We can update order save method to check for flag `_recalculate_tickets` and if True it would call the function `recalculate_ticket_price`. This way when creating new order (doesnt matter if from model or admin) we dont have to do the search for tickets and in ticket you can be sure that you update the price only for existing tickets (after save). I think we would end up with one extra if (but it seems easier to understand from code rather than have two super), and the recalculate_ticket_price will be in order where it makes more sense.

ricco386: Hey @simonsmiley what do you say about my suggestion? Would you be able to implement it? Please let me know either way. Thank you.

konfera/models/order.py
```diff
@@ -24,7 +24,7 @@
 
 
 class Order(KonferaModel):
-    price = models.DecimalField(decimal_places=2, max_digits=12, validators=[MinValueValidator(0)])
+    price = models.DecimalField(decimal_places=2, max_digits=12, validators=[MinValueValidator(0)], default=0)
```
ZoltanOnody: I don't like this default value. *Explicit is better than implicit* especially when we are talking about money. :-)

ricco386:
I am closing this PR.

I have create new PR #208 based on your work, but with changes I have suggested above. Thanks for your effort.

Issue #130, opened 2016-10-04
Create public tickets view


ricco386:
Add a view that will list all available ticket types, user will be able to select what kind of ticket he would like to register.

ricco386:
There is an issue #125 that changes the directory structure to avoid conflicts, lets do the parent issue first.

kapucko:
@ricco386 Do you want to modify the already existing volunteer registration form? It would be possible to add there dropdown menu with list of all available ticket types for given event. And the url could be /register/event/<event_slug> . That's my idea but maybe you had  something else in your mind.

ricco386:
@kapucko my idea was that we will have a page that will display all tickets, with the description, price, etc... So the user can choose, student or standard or supporter or anything else that will be defined as public ticket... Once user select ticket he will just click on the link type and on second page form will be presented where he fills the details.

So you can modify volunteer registration form, but instead of dropdown there will be hidden field. So far is seems to me that we will have only two form types, one for public tickets, that doesn have to have description field and private tickets, that will have description required. This seems to be good enought for version 0.2 in the near future we can have thoughts how to improove so you can set any field required per ticket type.

andela-ijubril:
Is anyone working on this?
I would like to give this a shot.

kapucko:
@andela-ijubril I haven't started yet, so you can take it.

ricco386:
@kapucko if you have time you can carry on with this issue, and @andela-ijubril can do the issue #120 Thanks guys...

kapucko:
Ok :)

kapucko:
In progress...

simonsmiley:
@kapucko are you still working on it?

kapucko:
Yes I'm

ricco386:
PR merged. thanks

PR #175, opened 2016-10-20
#130 list public tickets for event & public registration

kapucko:
There is code redundancy in public and private registration but I let there two views and extended url pattern to privateticket and publicticket....I'm not surewhat the difference between handlling private and publicregistration should be...

### DIFF
--- a/konfera/event/urls.py
+++ b/konfera/event/urls.py
@@ -11,5 +11,5 @@ urlpatterns = [
     url(r'^(?P<slug>[\w, -]+)/schedule/$', views.schedule_redirect, name='schedule'),
     url(r'^(?P<slug>[\w, -]+)/schedule/(?P<day>[\d, -]+)$', views.ScheduleView.as_view(), name='schedule'),
     url(r'^$', views.event_list, name='events'),
-    url(r'^(?P<slug>[\w, -]+)/tickets/$', views.event_public_tickets, name='event_speakers'),
+    url(r'^(?P<slug>[\w, -]+)/tickets/$', views.event_public_tickets, name='event_tickets'),
 ]
--- a/konfera/event/urls.py
+++ b/konfera/event/urls.py
@@ -11,4 +11,5 @@ urlpatterns = [
     url(r'^(?P<slug>[\w, -]+)/schedule/$', views.schedule_redirect, name='schedule'),
     url(r'^(?P<slug>[\w, -]+)/schedule/(?P<day>[\d, -]+)$', views.ScheduleView.as_view(), name='schedule'),
     url(r'^$', views.event_list, name='events'),
+    url(r'^(?P<slug>[\w, -]+)/tickets/$', views.event_public_tickets, name='event_speakers'),
 ]
--- a/konfera/event/views.py
+++ b/konfera/event/views.py
@@ -133,7 +133,7 @@ def event_public_tickets(request, slug):
     context['event'] = event
     available_tickets = event.tickettype_set.filter(accessibility=PUBLIC).exclude(attendee_type=AID)\
         .exclude(attendee_type=VOLUNTEER).exclude(attendee_type=PRESS)
-    available_tickets = [t for t in available_tickets if t.status==ACTIVE]
+    available_tickets = [t for t in available_tickets if t.status == ACTIVE]
     paginator = Paginator(available_tickets, 10)
     page = request.GET.get('page')
 
--- a/konfera/event/views.py
+++ b/konfera/event/views.py
@@ -9,7 +9,7 @@ from django.utils.translation import ugettext_lazy as _
 from konfera.event.forms import SpeakerForm, TalkForm
 from konfera.models.event import Event
 from konfera.models.talk import APPROVED, CFP
-from konfera.models.ticket_type import PUBLIC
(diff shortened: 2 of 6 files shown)


konfera/register/urls.py
```diff
@@ -19,7 +19,8 @@
 
 
 urlpatterns = [
-    url(r'^volunteer/event/(?P<slug>[\w, -]+)/$', views.register_volunteer, name='event_register_volunteer'),
-    url(r'^event/(?P<event_slug>[\w, -]+)/ticket/(?P<ticket_uuid>[\w, -]+)/$', views.private_registration,
+    url(r'^event/(?P<event_slug>[\w, -]+)/publicticket/(?P<ticket_uuid>[\w, -]+)/$', views.public_registration,
```
ZoltanOnody: we prefer `slug` over `event_slug`

konfera/templates/konfera/event_public_tickets.html
```diff
@@ -0,0 +1,41 @@
...
+            <a href="{% url 'event_public_registration' event.slug ticket.uuid %}" class="list-group-item">
+                <h4 class="list-group-item-heading"> {{ ticket.title }} </h4>
+                <p class="list-group-item-text">{{ ticket.description }}</p>
+                <p class="list-group-item-text">{% trans "Price" %}: {{ ticket.price }}</p>
+            </a>
+        {%  endfor %}
+    </div>
+    <nav aria-label="Page navigation">
+        <ul class="pagination">
+            {% if events.has_previous %}
```
ZoltanOnody: `events`? I think `tickets` should be here.

konfera/templates/konfera/event_public_tickets.html
```diff
@@ -0,0 +1,41 @@
...
+    <div class="list-group">
+        {% for ticket in tickets %}
+            <a href="{% url 'event_public_registration' event.slug ticket.uuid %}" class="list-group-item">
+                <h4 class="list-group-item-heading"> {{ ticket.title }} </h4>
+                <p class="list-group-item-text">{{ ticket.description }}</p>
+                <p class="list-group-item-text">{% trans "Price" %}: {{ ticket.price }}</p>
+            </a>
+        {%  endfor %}
+    </div>
+    <nav aria-label="Page navigation">
```
ricco386: How about we do not show the pagination if there is only one page?

konfera/event/views.py
```diff
@@ -123,3 +124,23 @@ def get_context_data(self, **kwargs):
...
 
         return context
+
+
+def event_public_tickets(request, slug):
+    context = dict()
+
+    event = get_object_or_404(Event.objects.published(), slug=slug)
+    context['event'] = event
+    available_tickets = event.tickettype_set.filter(accessibility=PUBLIC)
```
ricco386: Improve to filter also by ticket dates, so we do not dispaly expired tickets and also those that will by available in future.

konfera/register/views.py
```diff
@@ -9,13 +9,13 @@
 from konfera.register.forms import RegistrationForm
 
 
-def private_registration(request, event_slug, ticket_uuid):
+def private_registration(request, slug, ticket_uuid):
```
ricco386: private_registration and public_registration share a lot of same code base, how about we do one generic function for the stuff that is shared. We will have multiple view that will call this generic function with different parameters.

konfera/event/urls.py
```diff
@@ -11,4 +11,5 @@
     url(r'^(?P<slug>[\w, -]+)/schedule/$', views.schedule_redirect, name='schedule'),
     url(r'^(?P<slug>[\w, -]+)/schedule/(?P<day>[\d, -]+)$', views.ScheduleView.as_view(), name='schedule'),
     url(r'^$', views.event_list, name='events'),
+    url(r'^(?P<slug>[\w, -]+)/tickets/$', views.event_public_tickets, name='event_speakers'),
```
ricco386: i think more appropriate name would be event_tickets

konfera/register/urls.py
```diff
@@ -19,7 +19,8 @@
 
 
 urlpatterns = [
-    url(r'^volunteer/event/(?P<slug>[\w, -]+)/$', views.register_volunteer, name='event_register_volunteer'),
-    url(r'^event/(?P<event_slug>[\w, -]+)/ticket/(?P<ticket_uuid>[\w, -]+)/$', views.private_registration,
+    url(r'^event/(?P<slug>[\w, -]+)/publicticket/(?P<ticket_uuid>[\w, -]+)/$', views.public_registration,
```
ricco386: we dont need publicticket/privateticket in url as you have uuid which identifies ticket exactly and you can see in ticket what type it is.

My suggestion is to have  urls:
/event/<slug>/<uuid> - this way doesnt matter if public or private as uuid for private ones is not published anywhere on the site. We can send registration urls by email, etc. for private registrations.
/event/<slug>/volunteers - display volunteer form only if such ticket_type is available and is just one otherwise redirect to views.event_public_tickets
/event/<slug>/press  - display press form only if such ticket_type is available and is just one otherwise redirect to views.event_public_tickets
/event/<slug>/financial_aid  - display press form only if such ticket_type is available and is just one otherwise redirect to views.event_public_tickets

kapucko:
@ricco386 please check new commits, I hope that it'll meet your concept :)

kapucko:
thanks for feedback

Issue #173, opened 2016-10-20
Display order


ricco386:
Add new view that will display order details.

Once ticket is saved, redirect to show order details.

iraquitan:
Hi, I'd like to give it a try. Can you give me some initial details? The ticket is saved through the Orders admin view?

ricco386:
if you open in browser `http://127.0.0.1:8000/register/volunteer/event/2016/` you will see registration page. Once it is filled new ticket is created, ticket automatically generate an order. Purpose of this task is to create view that will display details of this order and update the view that user is redirected to see details of his order.

Order has uuid please use that in url of your new view. For example `/order/(?P<order_uuid>[\w, -]+)/` and it will display details of the order, and also related tickets. Please bear in mind that if order is created in admin it can have multiple tickets.

Hope this helps, if you have more questions, just ask. Thanks.

uj00007:
hey,i was looking into your issue ,but when i ran it ,it gives some error ,i think the apps are placed incorrectly in directories ,then i shifted them in example the it ran ok,the server was established but when i run the server in the browser,there are no pages configured or the opening html page or index.html ..is that ryt.?

ZoltanOnody:
closed in #179

PR #179, opened 2016-10-25
173 display order

iraquitan:
Added view to handle the order details. The order details template show the order info, receipt info and related tickets info in a table view. I included the view inside the event module views. Also added a custom filter to better handle currency in the template. I also had to make some minor changes [here](https://github.com/pyconsk/django-konfera/blob/master/konfera/register/views.py#L47) so I could have the ticket registration working. Any questions or improvements suggestions, just talk.


konfera/event/views.py
```diff
@@ -127,3 +128,16 @@ def event_public_tickets(request, slug):
...
+
+
+def event_order_detail(request, slug, order_uuid):
+    context = dict()
+    event = get_object_or_404(Event.objects.published(), slug=slug)
+    context['event'] = event
+    order = get_object_or_404(Order, uuid=order_uuid)
+    context['order'] = order
+
+    if order:
```
ZoltanOnody: This condition is unnecessary. `order` cannot be anything else than an object because a 404 would be raised.

konfera/templates/konfera/order_details.html
```diff
@@ -0,0 +1,93 @@
...
+{% trans "Order Detail" %}
+{% endblock %}
+
+{% block content %}
+  <div class="container">
+    <div class="row">
+      <div class="col-xs-12">
+        <div class="text-center">
+          <i class="fa fa-search-plus pull-left icon"></i>
+          <h2>Details for order #{{ order.variable_symbol }}</h2>
```
ZoltanOnody: Please use translation strings in the whole file.

konfera/templatetags/custom_filters.py
```diff
@@ -0,0 +1,19 @@
+from django import template
+import locale
+
+register = template.Library()
+
+# Use user's preferred locale
+user_locale = locale.setlocale(locale.LC_ALL, '')
+# Set default locale to en_US if user_locale is 'C' or None
```
ZoltanOnody: The code is self explanatory, there is no need for this comment.

konfera/templatetags/__init__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+"""
+ * Created by PyCharm.
```
ZoltanOnody: We don't store data like this anywhere else in the codebase so I'd remove.

iraquitan:
Added requested changes

konfera/event/views.py
```diff
@@ -127,3 +128,13 @@ def event_public_tickets(request, slug):
 
     context['tickets'] = available_tickets
     return render(request=request, template_name='konfera/event_public_tickets.html', context=context)
+
+
+def event_order_detail(request, slug, order_uuid):
+    context = dict()
+    event = get_object_or_404(Event.objects.published(), slug=slug)
```
ZoltanOnody: I just realized that we select `event` and `order` individually so user can change the `slug` and see an invalid order for another event. 

I suggest to remove the `slug` from the URL (and get the event object via `order.ticket.type.event`). (hmm.. but there is a problem!)

@ricco386 I think `Order` should have a `ForeignKey` to `Ticket`, currently it's vice versa, which IMO doesn't make sense.

konfera/templates/konfera/order_details.html
```diff
@@ -0,0 +1,92 @@
...
+            </div>
+          </div>
+          <div class="col-xs-12 col-md-6">
+            <div class="panel panel-info">
+              <div class="panel-heading">{% trans "Receipt" %}</div>
+              <div class="panel-body">
+                <strong>{% trans "Title: " %}</strong>{{ order.receipt_of.title }}<br>
+                <strong>{% trans "Amount: " %}</strong>{{ order.receipt_of.amount|currency }}<br>
+                <strong>{% trans "Street: " %}</strong>{{ order.receipt_of.street }}<br>
+                <strong>{% trans "Street2: " %}</strong>{{ order.receipt_of.street2 }}<br>
```
ZoltanOnody: I think we should show the optional fields only if they are filled in, what do you think?

konfera/event/views.py
iraquitan: I think it is because of what he said in an earlier comment in #179, that an order can have more than one ticket if added through the admin panel, so it is a one-to-many relationship and can have tickets from different events, so I think the info about the event only need to be shown in the ticket. And the slug really needs to be removed.

ZoltanOnody: Okay, then `order.ticket_set.all()[0].type.event` could work. But it's really ugly so maybe it would be a good idea to add method `event` (with property decorator) which would return the event for the order.

We also have to add validation so an order can have multiple tickets only if they are for the same event. (I will create a new issue for that.)

iraquitan: So orders can only have tickets from the same event? Because if an order can have tickets from multiple events, I thought of adding the event as a column in the related tickets table as in image below.

screenshot 2016-10-26 12 08 27

Another thing I was thinking is to put the label in the order status for better viewing, like in the previous and below images.

screenshot 2016-10-26 14 16 56

ricco386: @iraquitan great work the order looks very nice, and the label is great idea.

I think to in reality we will always have only tickets from the same event in one order, besides it occurs to me as a better solution. But somewhere on the receipt should the event be shown, doesnt have to be in ordered tickets section, it can be in order details, or somewhere where do you think is most relevant. 

@ZoltanOnody thanks for the review, and the logic how it will be used, I have explained [here](https://github.com/pyconsk/django-konfera/issues/173#issuecomment-255449895). I think the property decorator is the right thing to do.

konfera/event/urls.py
```diff
@@ -11,4 +11,5 @@
     url(r'^(?P<slug>[\w, -]+)/schedule/(?P<day>[\d, -]+)$', views.ScheduleView.as_view(), name='schedule'),
     url(r'^(?P<slug>[\w, -]+)/tickets/$', views.event_public_tickets, name='event_tickets'),
     url(r'^(?P<slug>[\w, -]+)/$', views.event_details_view, name='event_details'),
+    url(r'^order/(?P<order_uuid>[\w, -]+)/', views.event_order_detail, name='order_details'),
```
ZoltanOnody: Add `$` to the end of regex.

konfera/models/order.py
```diff
@@ -53,3 +53,10 @@ def to_pay(self):
     @property
     def variable_symbol(self):
         return str(int(self.uuid))[:10]
+
+    @property
+    def event(self):
+        if len(self.ticket_set.all()) > 0:
```
ZoltanOnody: This makes 2 queries to the DB. 

either "precache" the query by storing it in a variable (just 1 query to DB):
```
tickets = self.ticket_set.all()
if len(tickets):
    return tickets[0].type.event
```

or use 2 lightweight queries (in could be faster in this situation)
```
if self.ticket_set.exists():
    return self.ticket_set.first().type.event
```

konfera/models/order.py
```diff
@@ -53,3 +53,10 @@ def to_pay(self):
...
+            return self.ticket_set.all()[0].type.event
+        else:
```
ZoltanOnody: remove the `else` and `return None` directly in `event` ( = same behavior nicer code)

konfera/templates/konfera/order_details.html
```diff
@@ -0,0 +1,92 @@
...
+        <div class="row">
+          <div class="col-xs-12 col-md-6">
+            <div class="panel panel-default">
+              <div class="panel-heading">{% trans "Order Details" %}</div>
+              <div class="panel-body">
+                {% if order.event %}<strong>{% trans "Event: " %}</strong>{{ order.event }}<br>{% endif %}
+                {% if order.price %}<strong>{% trans "Price: " %}</strong>{{ order.price|currency }}<br>{% endif %}
+                {% if order.amount_paid %}<strong>{% trans "Amount paid: " %}</strong>{{ order.amount_paid|currency }}<br>{% endif %}
+                {% if order.discount %}<strong>{% trans "Discount: " %}</strong>{{ order.discount|currency }}<br>{% endif %}
+                {% if order.status %}<strong>{% trans "Status: " %}</strong><span class="label {{ status_label }}">{{ order.get_status_display }}</span><br>{% endif %}
```
ZoltanOnody: these lines are really long and they are hard to read, could you break them?

konfera/tests/test_templatetags.py
```diff
@@ -0,0 +1,34 @@
+from django.test import TestCase
+from ..templatetags.custom_filters import currency, set_user_locale
+import locale
+
+
+class TestCurrencyTag(TestCase):
+    def setUp(self):
+        self.value_str = ''
```
ZoltanOnody: From my experience it would be better to test a non-empty string.

konfera/tests/test_models.py
```diff
@@ -133,6 +133,27 @@ def test_left_to_pay2(self):
         entry = models.Order(price=155.5, discount=5.5, amount_paid=200)
         self.assertEqual(entry.left_to_pay, 0)
 
+    def test_order_event_with_ticket(self):
+        entry = models.Order.objects.create(price=155.5, discount=5.5)
+        # Create Event
+        title = random_string(16)
```
ZoltanOnody: Any reason why `title` should be a random string?

konfera/tests/test_models.py
```diff
@@ -133,6 +133,27 @@ def test_left_to_pay2(self):
...
+        slug = slugify(title)
+        date_to = timezone.now()
+        date_from = date_to + datetime.timedelta(days=1)
```
ZoltanOnody: Why is `date_from` after `date_to` ? 

Or a better question, but not directly to this PR . How is it possible it doesn't raise a `ValidationError` https://github.com/iraquitan/django-konfera/blob/59913d48efad30ee6a5f8de58b82ac67dbb66d9e/konfera/models/abstract.py#L33

@ricco386 I think `FromToModel`'s `.save` method needs to invoke `.clean()`, am I right?

iraquitan:
I'll check these tests in another computer.

konfera/register/views.py
```diff
@@ -44,7 +44,7 @@ def _register_ticket_attendee(request, slug, attendee_type):
     event = get_object_or_404(Event, slug=slug)
     ticket_types = TicketType.objects.filter(event=event, attendee_type=attendee_type)
 
-    if ticket_types and len(ticket_types) == 1 and ticket_types[0].status == TicketType.ACTIVE:
+    if ticket_types and len(ticket_types) == 1 and ticket_types[0].status == TicketType.STATUSES[TicketType.ACTIVE]:
```
ZoltanOnody: `.status` returns a human-readable text see https://github.com/pyconsk/django-konfera/blob/master/konfera/models/ticket_type.py#L76 . So that's why it evaluated to `False`.

iraquitan:
you can merge it

konfera/templatetags/custom_filters.py
```diff
@@ -0,0 +1,19 @@
+from django import template
+import locale
+
+register = template.Library()
+
+
+def set_default_locale():
+    user_locale = locale.setlocale(locale.LC_ALL, ('en_US', 'UTF-8'))
```
ZoltanOnody: `'en_US'` is fine as default but we should be able to change it in settings.

Issue #180, opened 2016-10-26
tickets must be for same event in order - validation


ZoltanOnody:
There can be multiple tickets in one order.
We need to validate that each ticket is for the same event.

HashCode55:
Hey, I'd take this one... Can you just give a gist what/where the work needs to be done?

DeKaN:
When you want to validate it?

ZoltanOnody:
Before saving the order.

DeKaN:
I will write code for it

DeKaN:
@ZoltanOnody You can't check this before saving order, only before saving ticket for order.

ZoltanOnody:
Okay, so do it that way if it works.

ricco386:
PR #181 has been merged long time ago, we just forgot to close the issue.

PR #181, opened 2016-10-26
Add check "tickets must be for same event in order" (#180)

DeKaN:


konfera/tests/test_models.py
```diff
@@ -135,21 +133,18 @@ def test_left_to_pay2(self):
...
 
 class ReceiptTest(TestCase):
-
     def test_string_representation(self):
         entry = models.Receipt(title="Test Receipt title")
         self.assertEqual(str(entry), entry.title)
 
 
 class RoomModelTest(TestCase):
-
```
ZoltanOnody: Those blank lines after `class` definition where there for readability purpose. I think we use it that way everywhere so let's keep them here as well (for consistency).

konfera/tests/test_models.py
```diff
@@ -276,6 +268,25 @@ def test_clean_and_save_ticket_discount_code(self):
                                                  discount_code=self.discount_code_2)
         self.assertRaises(ValidationError, ticket_with_invalid_code.save)
 
+    def test_save_tickets_for_different_event(self):
+        title1 = random_string(128)
```
ZoltanOnody: Personally I prefer when the title is static, the assertion messages are then more readable and it takes less time to run the tests.

DeKaN:
@ZoltanOnody Fixed it in new commit

ZoltanOnody:
Strange, I am still getting 500 error.

image

Does the code work for you?

Also change " ... for same event" to "... for **the** same event"

DeKaN:
@ZoltanOnody I've added validation for admin form.

konfera/admin.py
```diff
@@ -155,11 +156,23 @@ class LocationAdmin(admin.ModelAdmin):
 admin.site.register(Location, LocationAdmin)
 
 
+class OrderedTicketsInlineFormSet(forms.models.BaseInlineFormSet):
```
ZoltanOnody: Create a `forms.py` file and move this code there.

DeKaN:
@ZoltanOnody You press Save button in order form at admin area and:

1. admin form for `Order` calls `full_clean`
1. admin form for `Receipt` calls `full_clean` (**Receipt.order is None**)
1. admin form for each `Ticket` calls `full_clean` (**Ticket.order is None**)
1. all clean passed and form call `save `, but `Ticket.save` (**Ticket.order is not None** at this moment) calls `clean` without catching `ValidationError`

ZoltanOnody:
Thank you!

Issue #156, opened 2016-10-09
Update schedule model validation


ricco386:
When storing schedule model add clean and save methods (see ticket for inspiration), that will do the following validation:
- Only approved talk can be scheduled
- If talk is selected duration is copied from talk
- Event is related to location and location has room, make sure selected room in schedule belongs to Event's location
- Make sure date and time is within the event's range
- Make sure system does not allow store two events at the same time in the same room, eg. schedule datetime + duration in room is unique.

masterp4dev:
Hi @ricco386, I would like to try this.

ricco386:
Hi @masterp4dev sure go ahead...

ZoltanOnody:
Hey @masterp4dev how is it going? If you need any advice, feel free to ask.

EmilLuta:
Hello guys. I'm not sure on how the current progress is going, but I was wondering if I could give it a go. What do you think?

ZoltanOnody:
@EmilLuta sure :)

EmilLuta:
@ZoltanOnody Seems like I'll have to drop this one, sadly. Feel free to reassign it to someone else.

ricco386:
PR #212

ricco386:
we shall leave as it is and focus on #291

PR #182, opened 2016-10-27
added validation on schedule model

earvingemenez:
## Reference

https://github.com/pyconsk/django-konfera/issues/156

### DIFF
--- a/konfera/models/schedule.py
+++ b/konfera/models/schedule.py
@@ -1,5 +1,7 @@
 from django.core.validators import MinValueValidator, MaxValueValidator
+from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models.loading import get_model
 from django.utils.translation import ugettext_lazy as _
 
 from konfera.models.abstract import KonferaModel
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
 
     def __str__(self):
         return _('%(start)s (%(duration)s min)') % {'start': self.start, 'duration': self.duration}
+
+    def clean(self, *args, **kwargs):
+        # Only approved talk can be scheduled
+        if self.talk.status != 'approved':
+            raise ValidationError(_('You cannot schedule unapproved talks'))
+
+        # Event is related to location and location has room,
+        # make sure selected room in schedule belongs to Event's location
+        room = get_model('Room').objects.filter(location=event.location)
+        if location_room.first() != self.room:
+            raise ValidationError(_('The room that you have chosen is '
+                'not part of the scheduled room'))
+
+        # Make sure date and time is within the event's range
+
+        # Make sure system does not allow store two events at the same
+        # time in the same room, eg. schedule datetime + duration in room is unique.
+        schedules = get_model('Schedule').objects.filter(
+            start=self.start, duration=self.duration, room=self.room)
+        if schedules.exists():
+            # this means that there are events with the same schedule
+            raise ValidationError(_('Another event has the same schedule.'))
+
+        return super(Schedule, self).clean(*args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        if not self.id:
+            # If talk is selected duration is copied from talk
+            self.duration = self.talk.duration
+
+        return super(Schedule, self).save(*args, **kwargs)


konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
 
     def __str__(self):
         return _('%(start)s (%(duration)s min)') % {'start': self.start, 'duration': self.duration}
+
+    def clean(self, *args, **kwargs):
+        # Only approved talk can be scheduled
+        if self.talk.status != 'approved':
```
ZoltanOnody: Please import the `APPROVED` constant and compare the status with it.

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+            raise ValidationError(_('You cannot schedule unapproved talks'))
```
ZoltanOnody: Add a period to the end of the sentence.

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+
+        # Event is related to location and location has room,
+        # make sure selected room in schedule belongs to Event's location
+        room = get_model('Room').objects.filter(location=event.location)
```
ZoltanOnody: Is there any reason to use `get_model` over importing it directly using `from konfera.models import Room` ?

Same applies for the `get_model`s bellow.

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+        if location_room.first() != self.room:
+            raise ValidationError(_('The room that you have chosen is '
+                'not part of the scheduled room'))
```
ZoltanOnody: Add period to the end.

We have a 120 characters per line restriction, I think you can fit the whole sentence to one line. (but I didn't try it)

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+
+        # Make sure date and time is within the event's range
```
ZoltanOnody: Please implement it as well.

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+
+        # Make sure system does not allow store two events at the same
+        # time in the same room, eg. schedule datetime + duration in room is unique.
+        schedules = get_model('Schedule').objects.filter(
+            start=self.start, duration=self.duration, room=self.room)
+        if schedules.exists():
```
ZoltanOnody: What about when they are not fully overlapping?

For example:
schedule1 starts at 10:00 with duration 60minutes
schedule2 starts at 10:30 with duration 30minutes

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+            # this means that there are events with the same schedule
+            raise ValidationError(_('Another event has the same schedule.'))
```
ZoltanOnody: "The events in schedule cannot overlap if they are in the same room." is a more descriptive error

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+
+        return super(Schedule, self).clean(*args, **kwargs)
```
ZoltanOnody: We don't support python2 so use just `super().clean(*args, **kwargs)`.

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        if not self.id:
+            # If talk is selected duration is copied from talk
+            self.duration = self.talk.duration
+
+        return super(Schedule, self).save(*args, **kwargs)
```
ZoltanOnody: 1. `super().save(*args, **kwargs)`
1. add a newline at the end

konfera/models/schedule.py
```diff
@@ -23,3 +25,35 @@ class Schedule(KonferaModel):
...
+            # this means that there are events with the same schedule
+            raise ValidationError(_('Another event has the same schedule.'))
+
+        return super(Schedule, self).clean(*args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        if not self.id:
+            # If talk is selected duration is copied from talk
+            self.duration = self.talk.duration
```
ZoltanOnody: The comment says something else than the code actually does.

What is really going on in the code is:
If the schedule is being created  set the duration to the duration of the talk. (if no talk is selected it raises an error)

So `if self.talk` is the condition you need to write.

konfera/models/schedule.py
earvingemenez: @ZoltanOnody i read from the other Issue that the constant will be transfered from global to models (inside). so maybe i'll just wait for it to be merged then do the changes :)

konfera/models/schedule.py
earvingemenez: @ZoltanOnody actually there is no super valid reason. i commonly use that instead of adding an import line inside the method. also i was avoiding circular importing

konfera/models/schedule.py
ricco386: @earvingemenez it has been merged.

ZoltanOnody:
@earvingemenez are you planning to finish it?

ricco386:
I have updated @earvingemenez's code and added @ZoltanOnody comments in new PR #212

PR #183, opened 2016-10-28
applied validation on scheduled model

ralphleyga:
https://github.com/pyconsk/django-konfera/issues/156


ZoltanOnody:
Thanks,  but as you can see there is already a PR for this issue.

PR #189, opened 2016-11-01
Added settings for app

ricco386:
Added settings file for the app so we have way to define app settings with reasonable defaults, that is possible override in project settings

### DIFF
--- a/konfera/settings.py
+++ b/konfera/settings.py
@@ -1,0 +1,12 @@
+from django.utils.translation import ugettext_lazy as _
+from django.conf import settings
+
+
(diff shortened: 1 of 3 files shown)


konfera/settings.py
```diff
@@ -0,0 +1,12 @@
+from django.utils.translation import ugettext_lazy as _
+from django.conf import settings
+
+
+GOOGLE_ANALYTICS = getattr(settings, 'GOOGLE_ANALYTICS', None)  # just define analytics code: 'UA-XXXXXXXX-X'
+
+TALK_DURATION = getattr(settings, 'TALK_DURATION',
+                        (
```
ZoltanOnody: I don't like how this is formatted but it's pep8 valid so I suppose it's okay.

Issue #99, opened 2016-09-27
Add ability to process PayPal payments


ricco386:
PayPal has API that allows to create payments and notify us via callbacks:  https://developer.paypal.com/docs/integration/web/accept-paypal-payment/

Create a view where user can select PayPal Payment as a payment method for his order. Fill out necessary data, and we post data to PayPal, (user will need to confirm the payment) and on successful payment we are notified by PayPal, and order status changes to PAID.

Each order has to have its own unique UUID, and most probably we will communicate with this with PayPal.

More research to PayPal API needs to be done, for PayPal sandbox account talk to me (@RicCo386).

ricco386:
PR #191 has been merged. Thank you.

PR #191, opened 2016-11-03
Paypal payments

ZoltanOnody:
This is more or less a proof of concept. Issue #99.

todos:
- [x] todos in the code 
- [ ] tests

I created a payment view where users can see both bank account details (if they want to pay by transfer) and a paypal button. Example URL: `/order/fcdc461f-b0a9-47a6-bbea-32390fa6e888/payment`

image

When the user clicks to pay by paypal, he is redirected to the paypal page. After the payment he is redirected back to the order page. 
image

-----

Should we redirect from *payment view* to *order* if the order was already paid?


ZoltanOnody:
> how about we have paypal view as a separate file? Just a suggestion, but if we keep adding payment methods might be easier to manage.

IMHO it's not necessary for now, when we will add a new payment method we can do so but for now we have only 2 views and to have 2 files for 2 views seems a bit *strange*.

>  lets update order view to emit signal and each payment method can append its name and url where user should be directed (or possibly code to be rendered, eg. button, bank details). This way we can keep payments separate to konfera, but they can cooperate together. This way we wont need /order/uuid/payment view at all. But I presume this is work for separate issue, and we can have a more thought about it.

Let's move it to a separate issue.

payments/templates/payments/order_payment.html
```diff
@@ -0,0 +1,120 @@
...
+              <strong>{% trans "Bank name:" %}</strong> Fio banka, a. s., pobočka zahraničnej banky<br>
+              <strong>{% trans "Bank account EUR (SK):" %}</strong> 2300870269/8330<br>
+              <strong>{% trans "Bank account EUR (CZ):" %}</strong> 2300870269/2010<br>
+              <strong>{% trans "Bank account EUR (IBAN / international):" %}</strong> SK48 8330 0000 0023 0087 0269<br>
+              <strong>{% trans "SWIFT/BIC:" %}</strong> FIOZSKBAXXX
+            </p>
+
+            <h3>{% trans "Pay by PayPal" %}</h3>
+            <div class="alert alert-warning" role="alert">
+              {% trans "<strong>Warning!</strong> We charge additional 2% for PayPal payments." %}
```
ricco386: You have forgot to use setting value instead of hard coded constant.

payments/templates/payments/order_payment.html
```diff
@@ -0,0 +1,120 @@
...
+            </div>
+
+            <a href="{% url 'konfera_payments:paypal_button_url' order.uuid %}" class="btn btn-success">
+                {% trans "Pay by PayPal" %}
+            </a>
+          </div>
+        </div>
+        <hr>
+        <div class="row">
```
ricco386: I have a suggestion for this template. Lets use `{% extends 'konfera/order_details.html' %}` and in konfera/order_details.html add new block order_details_header (or something similar) and in this template instead of `{% block content %}` use `{% block order_details_header %}` this way you dont need all the code below, and if anything changes in konfera template will change here automatically as well. (additional benefit would be we can add setting for register redirect (default order_details), that can be set in local_settings to point here, without making konfera dependend on payments).

payments/templates/payments/order_payment.html
```diff
@@ -0,0 +1,120 @@
...
+                {% if order.amount_paid %}
+                  <strong>{% trans "Amount paid: " %}</strong>{{ order.amount_paid|currency }}<br>
+                {% endif %}
+                {% if order.discount %}
+                  <strong>{% trans "Discount: " %}</strong>{{ order.discount|currency }}<br>
+                {% endif %}
+                {% if order.status %}
+                  <strong>{% trans "Status: " %}</strong>
+                  <span class="label {{ status_label }}">
+                    {{ order.get_status_display }}
```
ricco386: doesnt seems to be working in current setup, but never mind if you implemented suggestion above.

payments/views.py
```diff
@@ -0,0 +1,151 @@
...
+    @staticmethod
+    def get_paypal_url(payment):
+        """ Return the PayPal URL where the user can pay for the order """
+        for link in payment.links:
+            if link.method == 'REDIRECT':
+                return str(link.href)
+
+        raise Exception("REDIRECT url not found when creating payment {id}".format(id=payment.id))
+
+    def pay(self, request, order):
```
ricco386: seems this can be static method

payments/views.py
```diff
@@ -0,0 +1,151 @@
...
+            except Exception as e:
+                logger.error(str(e))
+
+        logger.error("Payment for order(pk={order}) couldn't be created! Error: {err}".format(
+            order=order.pk, err=payment.error))
+        messages.error(request, _('Something went wrong, try again later.'))
+
+        return redirect('konfera_payments:payment_options', order_uuid=str(order.uuid))
+
+    def success(self, request, order):
```
ricco386: seems this can be static  method as well

payments/views.py
```diff
@@ -0,0 +1,151 @@
...
+        if order.status == Order.PAID:
+            return redirect('order_details', order_uuid=str(order.uuid))
+
+        status = request.GET.get('status')
+
+        if status not in ['success', 'failed']:
+            return self.pay(request, order)
+
+        if status == 'success' and self.success(request, order):
+            messages.success(request, _('Order successfully paid!'))
```
ricco386: Please add logging.

payments/views.py
```diff
@@ -0,0 +1,151 @@
...
+        else:
+            messages.error(request, _('Something went wrong, try again later.'))
```
ricco386: Please add logging.

payments/views.py
```diff
@@ -0,0 +1,151 @@
...
+        raise Exception("REDIRECT url not found when creating payment {id}".format(id=payment.id))
+
+    def pay(self, request, order):
+        """ Create the payment and redirect to PayPal or back to the order with an error message """
+        payment = paypalrestsdk.Payment({
+            "intent": "sale",
+            "payer": {"payment_method": "paypal"},
+            "redirect_urls": {
+                "return_url": request.build_absolute_uri().split('?')[0] + '?status=success',
+                "cancel_url": request.build_absolute_uri().split('?')[0] + '?status=failed',
```
ricco386: Just a question from curiosity (havent read the PayPal documentation), this are the only possibilities PayPal return? Eg. in cancel url, can you define more variables, that would be prepopulated byt PyaPal? To return extra info, for example if there is an error, what kind of error (insufficient funds, card expired, etc.) so we can show it to user, in get().

My assumption is that since user is on PayPal site, he sees all this error, and either finishes the purchase, or click on cancel (for ay reason, either error or he changes mind). In this case wouldnt be status canceled more descriptive? To failed is not users fault (problem might be somewhere else), but cancel is definte a users action.

payments/views.py
```diff
@@ -0,0 +1,151 @@
...
+            },
+            "transactions": [
+                {
+                    "amount": {
+                        "total": str(PayOrderByPaypal.get_paypal_price(order)),
+                        "currency": settings.PAYPAL_CURRENCY,
+                    },
+                    "description": _("Payment for order with variable symbol: {vs}".format(vs=order.variable_symbol))
```
ricco386: In order we have property event, why dont we add event name (possibly ticket type) here as well? Will be more user friendly if the person is looking at PayPal's history after some time...

payments/utils.py
```diff
@@ -4,12 +4,13 @@
 import requests
```
ricco386: I know it hasnt been change in this issue, but I dont think we have this in requirements...

payments/views.py
ZoltanOnody: @ricco386 logging is in `.success()`.

payments/utils.py
ZoltanOnody: Let's add it then there (in a separate PR).

payments/templates/payments/order_payment.html
ZoltanOnody: I deleted the redundant code, so we don't display order info in payment (for now), let's create a separate issue for it.

payments/templates/payments/order_payment.html
ZoltanOnody: I deleted it. Let's make it in a separate PR

payments/views.py
ZoltanOnody: I think there are only these 2 options.

Issue #187, opened 2016-11-01
Landing page settings


ricco386:
Landing page (index) should redirect to one of the pages: latest conference, latest meetup, list of conferences, list of meetups (all this views are already build).

Add setting LANDING_PAGE that will have pre defined const. 'latest_conference' and do the redirect (This is current behaviour). If no conference or meetup is defined in the DB, log warning message and redirect to list.

mimmon:
I will take this one.

ricco386:
Isse has been merged. No need for this issue

PR #193, opened 2016-11-04
#187 added landing page settings and handle redirecting on /index

mimmon:
We can discuss a way to handle LANDING_PAGE  - either hardcode all possibiities, or define rules to determine them dynamically. Current code suggests using default event latest or earliest, in case none is found, redirect to (empty) list of events of the same type. 
Tests TBD.

### DIFF
--- a/konfera/settings.py
+++ b/konfera/settings.py
@@ -11,5 +11,9 @@ TALK_DURATION = getattr(settings, 'TALK_DURATION',
                             (45, _('45 min')),
                         ))
 
-# LANDING PAGE should be something like latest_conference or earliest_meetup or similar combinations
+"""LANDING PAGE is a composite of two keywords: <timewise>_<event>
+<timewise> can be: latest or earliest
+<event> can be (at the moment): conference or meetup
+possible combinations: latest_conference (DEFAULT), latest_meetup, earliest_conference, earliest_meetup
+"""
 LANDING_PAGE = getattr(settings, 'LANDING_PAGE', 'latest_conference')
--- a/konfera/settings.py
+++ b/konfera/settings.py
@@ -10,3 +10,6 @@ TALK_DURATION = getattr(settings, 'TALK_DURATION',
                             (30, _('30 min')),
                             (45, _('45 min')),
                         ))
+
+# LANDING PAGE should be something like latest_conference or earliest_meetup or similar combinations
+LANDING_PAGE = getattr(settings, 'LANDING_PAGE', 'latest_conference')
--- a/konfera/tests/test_views.py
+++ b/konfera/tests/test_views.py
@@ -159,3 +159,49 @@ class TestOrderDetail(TestCase):
         response = self.client.get(reverse('order_details', kwargs={'order_uuid': self.order_await.uuid}))
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.context['status_label'], 'label-warning')
+
+
+class TestIndexRedirect(TestCase):
+
+    def setUp(self):
+        self.location = Location.objects.create(
+            title='FIIT', street='Ilkovicova', city='Bratislava', postcode='841 04', state='Slovakia', capacity=400,
+        )
+
+    def test_no_conference(self):
+        self.old_meetup = Event.objects.create(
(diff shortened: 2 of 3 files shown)


konfera/views.py
```diff
@@ -1,12 +1,28 @@
...
+        timewise, event_type = LANDING_PAGE.split('_')
+        timewise = timewise.lower()
+        event_type = event_type.upper()
+    except BaseException as be:
+        raise (_('%s\nIncorrect LANDING_PAGE setting, need latest|earliest_conference|meetup', be))
+
+    events = Event.objects.published().filter(event_type=getattr(Event, event_type))
+    selected_event = getattr(events, timewise)('date_from')
+    if not selected_event:
+        messages.info(request, _('No %s event has been found. Redirected to default list.' % event_type))
```
ZoltanOnody: Does `ugettext_lazy` work in views?

konfera/views.py
```diff
@@ -1,12 +1,28 @@
 from django.shortcuts import redirect
 from django.views.generic import ListView
+from django.shortcuts import render
```
ZoltanOnody: imports should be alphabetically

konfera/settings.py
```diff
@@ -10,3 +10,6 @@
                             (30, _('30 min')),
                             (45, _('45 min')),
                         ))
+
+# LANDING PAGE should be something like latest_conference or earliest_meetup or similar combinations
```
mimmon: There are four possibilities, I will specify them all.

mimmon:
I definitely want to write to tests for it. However, it will be my first view test, so I need some time to learn how to set up and perform this type of tests.

konfera/views.py
```diff
@@ -1,12 +1,28 @@
...
 
 def index(request):
-    latest_conference = Event.objects.published().filter(event_type=Event.CONFERENCE).latest('date_from')
-    return redirect('event_details', slug=latest_conference.slug)
+    try:
+        timewise, event_type = LANDING_PAGE.split('_')
+        timewise = timewise.lower()
+        event_type = event_type.upper()
+    except BaseException as be:
+        raise (_('%s\nIncorrect LANDING_PAGE setting, need latest|earliest_conference|meetup', be))
```
ricco386: I agree, no need for translations of error messages. This will end in log, which will be seen only by admin...

ricco386:
@mimmon currently there is no rush for this issue, as there is already functionality for redirect (also deployed  live). So I think you can take your time and write a tests. If you are stuck and need help just ask here.

konfera/views.py
```diff
@@ -14,12 +13,12 @@ def index(request):
...
     except BaseException as be:
-        raise (_('%s\nIncorrect LANDING_PAGE setting, need latest|earliest_conference|meetup', be))
+        raise ('%s\nIncorrect LANDING_PAGE setting, need latest|earliest_conference|meetup', be)
 
     events = Event.objects.published().filter(event_type=getattr(Event, event_type))
-    selected_event = getattr(events, timewise)('date_from')
+    selected_event = getattr(events, timewise)('date_from') if events else None
     if not selected_event:
-        messages.info(request, _('No %s event has been found. Redirected to default list.' % event_type))
+        messages.info(request, 'No %s event has been found. Redirected to default list.' % event_type)
```
ZoltanOnody: `messages` are shown to the user so they need to be translated.

konfera/tests/test_views.py
```diff
@@ -159,3 +159,49 @@ def test_order_status_await(self):
...
+class TestIndexRedirect(TestCase):
+
+    def setUp(self):
+        self.location = Location.objects.create(
+            title='FIIT', street='Ilkovicova', city='Bratislava', postcode='841 04', state='Slovakia', capacity=400,
+        )
+
+    def test_no_conference(self):
+        self.old_meetup = Event.objects.create(
+            title='Old meetup', slug='old-meetup', description='Old meetup', event_type='meetup',
```
ZoltanOnody: `event_type=Event.MEETUP`

same applies for the other tests

mimmon:
I unintendedly merged this issue, I suppose everything is correct according to previous requests, however, please check it so that no issue will arise. Sorry for inconvenience.

Issue #198, opened 2016-11-06
Add new fields to event model


ricco386:
Add new fields to event model (none of them is required): 
* *cfp_allowed*: Boolean field, default: True, help text: Weather to allow submitting talk proposals.
* *cfp_end*: Datetime field, help text: Call for proposals deadline.
* *contact_email*: EmailField, help text: Publicly displayed email to contact organizers.
* *coc*: TextField, name: Code of Condict
* *coc_phone*: CharField, help text: Publicly displayed phone to contact organizers.
* *coc_phone2*: CharField, help text: Publicly displayed phone to contact organizers.
 
Add validation for the cfp_end field have to be before date_from and required if cfp_allowed.
Update existing cfp view to show 404 if cfp_allowed is False. Once we pass the cfp_end date do not show (allow to submit) form, and display message (Something along this): 

> Thank you for your interest, but unfortunately call for proposals deadline was at: `cfp_end`. However if you feel like you have to submit your talk anyway, feel free to contact us: `contact_email`

Update admin area to be able to update new fields and create a migration for the new fields.

mimmon:
I will take this, if it is not urgent. Please let me know so that I can assign it to myself.

ricco386:
Sure, do it...

ricco386:
@mimmon I have updated issue and moved few fields (coc) into separate issue, as we need it on the site. When I created that bug I totally forgot you already have it here. Sorry about this.

PR #202, opened 2016-11-13
Issue198

mimmon:


konfera/models/event.py
```diff
@@ -41,6 +44,20 @@ class Event(FromToModel):
     footer_text = models.TextField(blank=True)
     analytics = models.TextField(blank=True)
 
+    cfp_allowed = models.BooleanField(default=True, help_text=_('Is it allowed to submit talk proposals?'))
+    cfp_end = models.DateTimeField(verbose_name=_('CFP deadline'),
+                                   default=datetime(2000, 1, 1, 0, 0, tzinfo=timezone.utc),
```
ZoltanOnody: Why don't we use `null=True` instead of some arbitrary default value?

konfera/models/event.py
```diff
@@ -41,6 +44,20 @@ class Event(FromToModel):
...
+                                   help_text=_('Call for proposals deadline.'))
```
ZoltanOnody: Is the `help_text` necessary? It says the same as the `verbose_name`.

konfera/models/event.py
```diff
@@ -53,6 +70,17 @@ def __str__(self):
     def duration(self):
         return (self.date_to - self.date_from).days + 1
 
+    @property
+    def cfp_open(self):
+        return self.cfp_allowed and self.cfp_end >= datetime.now(tz=timezone.utc)
```
ZoltanOnody: I think `timezone.now()` does the same as `datetime.now(tz=timezone.utc)` isn't it?

konfera/models/event.py
```diff
@@ -53,6 +70,17 @@ def __str__(self):
...
+
+    def clean(self):
+        if self.cfp_allowed and self.cfp_end == datetime(2000, 1, 1, 0, 0, tzinfo=timezone.utc):
```
ZoltanOnody: If we change the model to be `blank=False, null=True` we can change the condition to `self.cfp_allowed and not self.cfp_end` which is more readable.

konfera/templates/konfera/event/cfp_form.html
```diff
@@ -2,17 +2,31 @@
 {% load i18n %}
 
 {% block event_details %}
+  {% block top_cfp %}{% endblock %}
```
ZoltanOnody: This line should be indented the same way as the `if` bellow.

konfera/templates/konfera/event/cfp_form.html
```diff
@@ -2,17 +2,31 @@
...
-        <span class="glyphicon glyphicon-thumbs-up" aria-hidden="true"></span> {% trans "Submit proposal" %}
-      </button>
-    </form>
-    {% block bottom_cfp %}{% endblock %}
-
+    {% else %}
+      <div class="row">
+        <div class="col-md-8">
+          <p>
+            {% trans "Thank you for your interest, but unfortunately call for proposals deadline was at: "
```
ZoltanOnody: You are missing `%}` here.

konfera/templates/konfera/event/cfp_form.html
```diff
@@ -2,17 +2,31 @@
...
+            {{ event.cfp_end }}.
+            <br/>
+            {% trans "However, if you feel like you have to submit your talk anyway, feel free to contact us:"
```
ZoltanOnody: `%}`

konfera/templates/konfera/event/cfp_form.html
```diff
@@ -2,17 +2,31 @@
...
+            {{ event.contact_email }}
+            <br/>
```
ZoltanOnody: Using `<br/>` for *design* purposes is an antipattern. If it looks strange without it, please update the *css* files.

konfera/tests/test_models.py
```diff
@@ -94,6 +94,18 @@ def test_slug(self):
...
+        now = datetime.datetime.now()
+        tomorrow = now + datetime.timedelta(days=1)
+        today = datetime.date.today()
+        intwodays = today + datetime.timedelta(days=2)
+        event = Event(title='Test event', date_from=today, date_to=intwodays, cfp_allowed=True)
+        self.assertRaises(ValidationError, event.clean)
+        event1 = Event(title='Test event', date_from=today, date_to=intwodays, cfp_allowed=True, cfp_end=now)
+        self.assertRaises(ValidationError, event1.clean)
+        Event(title='Test normal event', date_from=intwodays, date_to=intwodays, cfp_end=tomorrow)
+        self.assert_(True)
```
ZoltanOnody: What does this line do?

konfera/tests/test_views.py
```diff
@@ -81,6 +92,16 @@ def test_cfp_non_existing_event(self):
...
 
+    def test_cfp_not_allowed(self):
+        url = reverse('event_cfp_form', kwargs={'slug': 'cfp-not-allowed'})
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 404)
+
+    def test_cfp_after_deadline(self):
+        url = reverse('event_cfp_form', kwargs={'slug': 'passed-cfp'})
+        response = self.client.get(url)
+        self.assertIn('{% trans "Thank you for your interest, but unfortunately', str(response.content))
```
ZoltanOnody: `{% trans "` should not be in the output HTML file, if it's there it means something went wrong. (in this case you forgot the closing `%}` in the templates)

konfera/models/event.py
mimmon: this was the original intention, but during the tests it required default value, I did not realize I can easily use null=True to make it work

konfera/tests/test_views.py
mimmon: I fixed it, but I wonder if it is possible to test it when the translation is applied and how...

ZoltanOnody: It is not necessary to test whether the translation was applied or not. 

But you could do it with https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#explicitly-setting-the-active-language . You activate a different language for example `sk` and then check if the page shows the text in Slovak.

konfera/tests/test_models.py
ZoltanOnody: What about 
```
event2 = Event(title='Test normal event', date_from=intwodays, date_to=intwodays, cfp_end=tomorrow)
self.assertIsNone(event2.clean())
```
I think it could do the job. :)

mimmon: Ok, I will test it this way. During the tests I found that the abstract FromTo model required date_start to be less then date_end. However, we can have an event that only lasts one day so I changed the logic of validator from ">=" to ">". In case it violates the logic or intention, please let me know so that I can revert it and reset the tests.

mimmon:
I think it is now ready. However, don't we need to create/update view where CoC will be displayed?

ricco386:
Please dont merge your pull requests until at least one person review it and approve. This is the rule for everybody.

Yes create a view apply similar logic that was used in issue #198

Issue #197, opened 2016-11-06
Add venue page


ricco386:
If we have event location set and location has filled in field how to get here, event should have accessible url `/<event>/venue` and this page will display content of the how to get here section. Make sure template do not escape html characters.

kubco2:
Im working on it.

ricco386:
PR has been merged. Thanks.

PR #206, opened 2016-11-14
Issue 197 event venue view

kubco2:
Related to https://github.com/pyconsk/django-konfera/issues/197 
- added event venue view
- tests for view

### DIFF
--- a/konfera/event/urls.py
+++ b/konfera/event/urls.py
@@ -4,6 +4,7 @@ from konfera.event import views
 
 
 urlpatterns = [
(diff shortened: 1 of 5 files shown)


kubco2:
I fixed PEP8 issue about my files, however there are errors in admin.py which is not part of this pull request, should it be fixed here?

ricco386:
Thank you for your effort I will merge this PR.

No need to fix it here, since you didnt introduce this errors. However if you could please create new PR with fixes.

PR #207, opened 2016-11-14
fix PEP8 issue in admin.py

kubco2:
Fix broken PEP8 tests from https://github.com/pyconsk/django-konfera/pull/206


ricco386:
Thanks for the PR.

PR #208, opened 2016-11-15
Calculate price and discount

ricco386:
I have just updated last bit from PR #170 that was missing so we can close issue #149


konfera/models/order.py
```diff
@@ -23,7 +23,7 @@ class Order(KonferaModel):
         (CANCELLED, _('Cancelled')),
     )
 
-    price = models.DecimalField(decimal_places=2, max_digits=12, validators=[MinValueValidator(0)])
+    price = models.DecimalField(decimal_places=2, max_digits=12, validators=[MinValueValidator(0)], default=0)
```
ZoltanOnody: This will need a migration.

Issue #194, opened 2016-11-04
bootstrap alert messages


ZoltanOnody:
In https://github.com/pyconsk/django-konfera/blob/master/konfera/templates/konfera/base.html#L28 there is `alert-success` hardcoded so if we add an error message as `messages.error(..)`  it shows up green.

mimmon:
I'll take this

ricco386:
PR has been merged. Thanks.

PR #209, opened 2016-11-15
#194 hardcoded alert-success changed to message.tags, tests TBD

mimmon:
I am struggling for several days with writing the tests for this. I added two possible ways to generate error alert, but none of them worked. 
I was also unable to generate the alert manually in the application. Please let me know if you have an idea how to tests it.


ZoltanOnody:
Have you tried http://stackoverflow.com/questions/2897609/how-can-i-unit-test-django-messages/14909727#14909727 ?

mimmon:
I tried, but I cannot add alert message as this requires a request. The
linked approach already has message in response. Moreover I would like to
test not the message itself, but the alert's class (alert-error, or
probably also alert-success).
I also tried another approach with generating request via Factory, this was
unsuccessul because django required to install middleware which was
declared in example/settings.py

ZoltanOnody:
I'll have a look on it tomorrow :)

ZoltanOnody:
After 2 hours of struggling I am giving up. 

I had a closer look on the issue and I realized even if we could test it we would be just testing constants from the settings.

I am sincerely sorry you were wasting you time because of my bad advice on the meetup.

mimmon:
No problem, I think that the more I work with the project the more I learn, so it was definitely not wasted time.

konfera/templates/konfera/base.html
```diff
@@ -26,7 +26,8 @@
 
   {% if messages %}
     {% for message in messages %}
-    <div class="alert alert-success alert-dismissible{% if message.tags %} {{ message.tags }}{% endif %}" role="alert">
+    <div class="alert alert-{{ message.tags }} alert-dismissible{% if message.tags %} {{ message.tags }}{% endif %}"
```
ZoltanOnody: We don't need `{% if message.tags %} {{ message.tags }}{% endif %}` do we? 
I mean when a tag is set it shows up as `alert-XYZ` which is fine for our design.

Issue #204, opened 2016-11-13
Add new model Organizer


ricco386:
Add new model organizer, model should have following fields:

* title - char 128 (required)
* street - char 128
* street2 - char 128
* city - char 128
* postcode - char 12
* state  - choice COUNTRIES (models.countries)
* comapnyid - char 32
* taxid - char 32
* vatid - char 32
* about_us

make a migration and form for admin to be able to modify all of this information.

Update event model to have foreign key to organizer, e.g. we should be able to set organizer for event.

mimmon:
I am aware that my last pull request is not yet finished because of missing tests, however, I will need a little help there. Meanwhile, I would like to take this one. Is it possible?

ricco386:
sure, if you have a PR on review it is OK to start another. What I meant is not to start 3-4 and none of them have on review.

ricco386:
PR has been merged. Thanks

PR #210, opened 2016-11-17
Issue204

mimmon:
#204


konfera/migrations/0010_auto_20161117_0652.py
```diff
@@ -0,0 +1,44 @@
...
+                ('about_us', models.TextField(blank=True)),
+            ],
+            options={
+                'abstract': False,
+            },
+        ),
+        migrations.AddField(
+            model_name='event',
+            name='organizer',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='organized_events', to='konfera.Organizer'),
```
ZoltanOnody: `SET_NULL` would be better choice than `CASCADE`.

@ricco386 I think we should check every foreign key in the whole project and make sure `on_delete` works as expected.

konfera/models/organizer.py
```diff
@@ -0,0 +1,24 @@
+from konfera.models.countries import COUNTRIES
+from django.db import models
```
ZoltanOnody: `django.db` should be the first import and `.abstract` should be before `.countries`

konfera/models/organizer.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Organizer(KonferaModel):
+    COUNTRY_DEFAULT = 'SK'
+
+    title = models.CharField(max_length=128)
+    street = models.CharField(max_length=128)
+    street2 = models.CharField(max_length=128)
+    city = models.CharField(max_length=128)
+    postcode = models.CharField(max_length=12)
+    country = models.CharField(
```
ZoltanOnody: Can't we fit it in a single line?

konfera/tests/test_models.py
```diff
@@ -409,3 +409,15 @@ def test_status(self):
...
+
+class OrganizerTest(TestCase):
+
+    def setUp(self):
+        self.first_organizer = models.Organizer(title='Mysterious Organizer', street='1 Up street', city='Big City',
+                                                company_id='123',
+                                                about_us="World famous yet unknown conference organizer.")
+
+    def test_save(self):
+        self.assertTrue(self.first_organizer.save)
```
ZoltanOnody: hmm... `.save` is a method so you have to invoke it, otherwise the assertion doesn't make any sense see:

```
>>> def f(): return False
>>> bool(f)
True
```

If you called `.save()` you would see it throws an error because `street2` is a mandatory field.

konfera/models/organizer.py
```diff
@@ -0,0 +1,24 @@
+from konfera.models.countries import COUNTRIES
+from django.db import models
+from konfera.models.abstract import KonferaModel
+
+
+class Organizer(KonferaModel):
```
ZoltanOnody: I think @ricco386 meant that only the `title` should be required and the other fields should have the `blank=True` attribute.

konfera/models/organizer.py
mimmon: There are also company/organization details like ID, tax ID or VAT ID so I suppose we want to keep all the address details.

konfera/models/organizer.py
mimmon: I have not realized fields are required unless defined otherwise. I will change it. The question is whether the about_us field should not be required.

konfera/tests/test_models.py
mimmon: Thank you for your feedback, I will commit the changes tonight.

konfera/models/organizer.py
mimmon: I see, I will change it, it is 89 chars, so pep8 linter might be merciful :)

konfera/tests/test_models.py
ZoltanOnody: You are welcome :)

konfera/models/event.py
```diff
@@ -40,6 +40,7 @@ class Event(FromToModel):
     event_type = models.CharField(choices=EVENT_TYPE_CHOICES, max_length=20)
     status = models.CharField(choices=EVENT_STATUS_CHOICES, max_length=20)
     location = models.ForeignKey('Location', related_name='events')
+    organizer = models.ForeignKey('Organizer', related_name='organized_events', null=True)
```
ZoltanOnody: It looks like you changed the `on_delete` by hand in migration. It would work but we plan to remove all the migrations and create a new one(so we don't have 10 files in *django-konfera 0.2*) so the `on_delete` would be overwritten.

We should set `SET_NULL` directly in models (see https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey ) so the models corresponds with the migrations.

konfera/tests/test_models.py
```diff
@@ -409,3 +409,17 @@ def test_status(self):
...
+
+    def setUp(self):
+        self.first_organizer = models.Organizer(title='Mysterious Organizer', street='1 Up street', city='Big City',
+                                                company_id='123',
+                                                about_us="World famous yet unknown conference organizer.")
+        self.noname_organizer = models.Organizer(street='2 Random', city='Small City')
+
+    def test_save(self):
+        self.assertEquals(self.first_organizer.save(), None)
+        self.assertEquals(str(self.first_organizer), 'Mysterious Organizer')
```
ZoltanOnody: The test's name is `test_save` so we should test only the `.save()` method here. Please move this (`str`) assertion to a new test.

konfera/tests/test_models.py
```diff
@@ -409,3 +409,17 @@ def test_status(self):
...
         self.assertEquals(tt.status, TicketType.STATUSES[TicketType.EXPIRED])
+
+
+class OrganizerTest(TestCase):
+
+    def setUp(self):
+        self.first_organizer = models.Organizer(title='Mysterious Organizer', street='1 Up street', city='Big City',
+                                                company_id='123',
+                                                about_us="World famous yet unknown conference organizer.")
+        self.noname_organizer = models.Organizer(street='2 Random', city='Small City')
```
ZoltanOnody: You created the `noname_organizer`` to test it cannot be saved. So I think the only place we will use it is in `test_save` so please move it to that method as a local variable.

PR #212, opened 2016-11-19
Schedule/validation PR comments from #182

ricco386:
Updated work made in PR #182 so we can close it (I didnt implement missing bit, just fixed comments so it can go t o current release).

How ever I have a thought and it is not very practical to set schedule like this. We should probably generate it. I will create a new issue for the auto generation, we have to thought about it a bit, since we need:
* easily define talks order
* easily re-schedule
* easily cancel talk

I dont think current setup allows any of this (validation, event if I would complete  missing parts, is progress but to do this in admin is pain).

### DIFF
--- a/konfera/models/schedule.py
+++ b/konfera/models/schedule.py
@@ -29,12 +29,14 @@ class Schedule(KonferaModel):
         return _('%(start)s (%(duration)s min)') % {'start': self.start, 'duration': self.duration}
 
     def clean(self, *args, **kwargs):
+        duration_delta = datetime.timedelta(minutes=self.duration)
+
         # Only approved talk can be scheduled
         if self.talk and self.talk.status != self.talk.APPROVED:
             raise ValidationError({'talk': _('You cannot schedule unapproved talks.')})
 
         # Make sure date and time is within the event's range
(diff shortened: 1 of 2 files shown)


konfera/models/schedule.py
```diff
@@ -18,8 +22,41 @@ class Schedule(KonferaModel):
...
     def __str__(self):
         return _('%(start)s (%(duration)s min)') % {'start': self.start, 'duration': self.duration}
+
+    def clean(self, *args, **kwargs):
+        # Only approved talk can be scheduled
+        if self.talk and self.talk.status != self.talk.APPROVED:
+            raise ValidationError({'talk': _('You cannot schedule unapproved talks.')})
+
+        # Make sure date and time is within the event's range
+        if self.event.date_from > self.start or self.start > self.event.date_to:
```
ZoltanOnody: `self.start > self.event.date_to`

Shouldn't it be `self.start + datetime.timedelta(minutes=self.duration) > self.event.date_to` ? 

Otherwise it will be possible to set a schedule, for a 40 minute long talk, to 1 minute before the event ends.

PR #215, opened 2016-11-20
Added notification email once user register ticket

ricco386:


konfera/settings.py
```diff
@@ -11,9 +11,31 @@
...
 """
 LANDING_PAGE = getattr(settings, 'LANDING_PAGE', 'latest_conference')
+
+"""
+Register email notification.
+"""
+REGISTER_EMAIL_NOTIFY = getattr(settings, 'REGISTER_EMAIL_NOTIFY', False)
+REGISTER_EMAIL_BCC = getattr(settings, 'REGISTER_EMAIL_BCC', [])
+REGISTER_EMAIL = getattr(settings, 'REGISTER_EMAIL',
+                         'Dear {first_name} {last_name},\n\n' \
```
ZoltanOnody: I'd rather have this in a separate file (`register/email.txt` ?) than in the settings.

konfera/register/views.py
```diff
@@ -23,7 +28,30 @@ def _register_ticket(request, event, ticket_type):
...
         new_ticket.save()
 
-        messages.success(request, _('Thanks for registering...'))
+        if settings.REGISTER_EMAIL_NOTIFY:
+            messages.success(request, _('Thank you for ordering ticket. You will receive confirmation email soon.'))
+
+            event = new_ticket.type.event
+            order_url = request.build_absolute_uri(reverse('order_details', args=[new_ticket.order.uuid]))
+            event_url = request.build_absolute_uri(reverse('event_details', args=[event.slug]))
+            subject = 'Your ticket for {event}.'.format(event=new_ticket.type.event.title)
```
ZoltanOnody: - Translation strings.
- `event.title`

konfera/settings.py
ricco386: I was thinking about it, and it is better idea to have it in file (or in DB), as I assume we will have more mail templates and some of them might have longer text. However the difficult thing is how to do it in a way that it can be easily overloaded. I imagine that someone do pip install django-konfera (once it is there :)), and the question here is how do we let admin modify the template? I dont like the idea that he would edit file in installed app, neither he will have fake app loaded before ours to load the overloaded templates first, we might look in project for specific dir if it exists and if the file is there, or the cleanest way that occured to me was to create a DB schema that will prepopulate the templates and user can edit them in admin area, but seemd like a lot of work (since we want launch tickets this week), therefore i went with setting as easy way since you can do local setting ad have your modifications there. My suggestion is let have a wider discussion about it and create a separate issue for this? If you have better suggestion, please let me know as this should be super easy for app admin to use.

ricco386: I have create issue #225 for this, so we can merge and close this PR.

PR #222, opened 2016-11-23
Ticket admin fix

mimmon:
Removed description field from admin Add ticket form


ricco386:
I guess this was due to our chat with regards description field shown when registering ticket, but I probably didnt explain correctly what was my intention. I have already made those changes in PR #224

PR #224, opened 2016-11-24
Updated ticket to display proper fields when registering ticket

ricco386:
Refactored function for ticket register


mimmon:
LGTM

PR #227, opened 2016-11-24
Refactored templates

ricco386:
Refactored templates to use base template and moved event templates into event directory


mimmon:
LGTM

Issue #225, opened 2016-11-24
Email templates handling


ricco386:
Add ability to modify email templates easily. Currently we do have email template in settings.

Lets add new table, email_templates with following field:
name
template
html_template
counter

This way admin can easily modify email template, lets have a fixture that will pre-define the email templates. Existing email should be in this fixture. Update the current code to use this new feature.

mimmon:
I'll do it.

ricco386:
PR #229 merged. Thank you.

PR #229, opened 2016-11-26
#225 handling email templates

mimmon:
new model EmailTemplate
register email template loaded into database during migration
admin section
tests


konfera/admin.py
```diff
@@ -321,3 +321,25 @@ class OrganizerAdmin(admin.ModelAdmin):
 
 
 admin.site.register(Organizer, OrganizerAdmin)
+
+
+class EmailTemplateAdmin(admin.ModelAdmin):
+    list_display = ('name', 'text_template', 'html_template', 'counter')
```
ricco386: we dont need to see html_template here, I am not even sure about the text_template how it would look it the text is too long (which in mail might be). So let stay with text_template only for now.

konfera/admin.py
```diff
@@ -321,3 +321,25 @@ class OrganizerAdmin(admin.ModelAdmin):
...
+    readonly_fields = ('date_created', 'date_modified', 'counter')
+    fieldsets = (
+        (_('Basic Info'), {
+            'fields': ('name',),
+        }),
+        (_('Templates'), {
+            'fields': ('text_template', 'html_template',),
+        }),
+        (_('Modifications'), {
+            'fields': ('counter', 'date_created', 'date_modified',),
```
ricco386: Move the counter to basic info, so admin see it immediatelly

PR #231, opened 2016-11-26
Updated missing requirement

ricco386:


mimmon:
my virtualenv shows that (supposedly) konfera already installed requests==2.12.1

Issue #226, opened 2016-11-24
Add email notification after submitting CfP


ricco386:
Add email notification after submitting talk proposal. Have a look register/views.py how we do email notification when ticket is issued, add similar one for speaker who is proposing talk.

ricco386:
PR #233 has been merged. Thank you.

PR #233, opened 2016-11-27
#226 Notification after talk proposal 

mimmon:
email notification after submission
updated tests
notification template in migration


konfera/event/views.py
```diff
@@ -1,21 +1,35 @@
+import logging
+
 from datetime import timedelta
+from smtplib import SMTPException
 
+from django import VERSION
+from django.conf import settings as django_settings
```
ricco386: we dont need this import, use settings from konfera, you already have that import down below...

konfera/event/views.py
```diff
@@ -74,15 +88,50 @@ def dispatch(self, *args, **kwargs):
 
     def post(self, *args, **kwargs):
         context = self.get_context_data(**kwargs)
+        notify = getattr(django_settings, 'PROPOSAL_NOTIFY', getattr(settings, 'PROPOSAL_NOTIFY'))
```
ricco386: in konfera settings you are already getting django settings (this includes also local_settings) in there is PROPOSAL_NOTIFY set, with False as a default. No need to do it here again.

konfera/event/views.py
```diff
@@ -74,15 +88,50 @@ def dispatch(self, *args, **kwargs):
...
+            speaker = context['speaker_form'].save()
             talk_instance = context['talk_form'].save(commit=False)
-            talk_instance.primary_speaker = speaker_instance
+            talk_instance.primary_speaker = speaker
             talk_instance.event = context['event']
             talk_instance.status = talk_instance.status or Talk.CFP
             talk_instance.save()
-            messages.success(self.request, self.message_text)
+
+            if notify:
```
ricco386: You can do just:

`if settings.PROPOSAL_NOTIFY:`

as I dont see notify anywhere else.

konfera/settings.py
```diff
@@ -29,3 +29,6 @@
 """
 REGISTER_EMAIL_NOTIFY = getattr(settings, 'REGISTER_EMAIL_NOTIFY', False)
 REGISTER_EMAIL_BCC = getattr(settings, 'REGISTER_EMAIL_BCC', [])
+
+"""Notify after submitting proposal"""
+PROPOSAL_NOTIFY = getattr(settings, 'PROPOSAL_NOTIFY', False)
```
ricco386: Lets use generic variable naming, I suggest `PROPOSAL_EMAIL_NOTIFY` so it matches `REGISTER_EMAIL_NOTIFY`, eg. all other notification should be `SOMETHING_EMAIL_NOTIFY`

konfera/event/views.py
```diff
@@ -74,15 +88,50 @@ def dispatch(self, *args, **kwargs):
...
+                                 'event': self.event.title,
+                                 'talk': talk_instance.title,
+                                 'event_url': event_url,
+                                 'edit_url': edit_url,
+                                 'end_call': self.event.cfp_end}
+                text_content = template.text_template.format(**template_data)
+                html_content = template.html_template.format(**template_data)
+
+                msg = EmailMultiAlternatives(subject, text_content, to=[speaker.email],
+                                             bcc=settings.REGISTER_EMAIL_BCC)
```
ricco386: I dont like mixing of settings in multiple views, so either rename to `settings.EMAIL_NOTIFY_BCC` to be generic and used in all future email notifications, or add new setting `PROPOSAL_EMAIL_BCC`. I think I prefer the first one (generic name).

konfera/migrations/0014_auto_20161127_0812.py
```diff
@@ -0,0 +1,42 @@
...
+from django.db import migrations
+
+
+def forwards_func(apps, schema_editor):
+    EmailTemplate = apps.get_model("konfera", "EmailTemplate")
+    db_alias = schema_editor.connection.alias
+    text = """Dear {first_name} {last_name},\n\n
+              thank you for submitting proposal for {event}.\nYou can edit your submission here: {edit_url}.\n\n
+              The call for proposals ends at {end_call}. We will inform you about the acceptance status and further
+              details after this date.\n\n\n
```
ricco386: How about we change:

> We will inform you about the acceptance status and further details after this date.
 
to

> We will inform you about the acceptance status and further details shortly after this date.

konfera/event/views.py
mimmon: django_settings are there because otherwise I was unable to test behaviour under different settings

konfera/event/views.py
mimmon: I thought BCC can be universal for all notifications, but you are right - I will make a new bcc for this  notifications

konfera/event/views.py
ZoltanOnody: Yes you can. See https://github.com/pyconsk/django-konfera/blob/master/payments/tests/test_views.py#L27 .

konfera/event/views.py
ricco386: I meant universal BBC, just rename the existing one to: EMAIL_NOTIFY_BCC everywhere...

PR #237, opened 2016-11-28
Quick workaround not do tispay dollards for all

ricco386:


ZoltanOnody:
I think i tested the previous code and it was working correctly. Did you set locale in settings?

PR #240, opened 2016-11-28
updated currency to be handled in more transparent way

ricco386:
Fix broken Travis test


ricco386:
if you ask about templatetag than nowhere, just prepared to use if anyone wants it. It you asked about setting value, that  we do send it to PayPal.

PR #242, opened 2016-11-28
Updated fio bank management command

ricco386:


payments/utils.py
```diff
@@ -25,7 +25,7 @@ def _get_last_payments():
     client = FioBank(token=settings.FIO_BANK_TOKEN)
 
     today = timezone.now()
-    date_from = (today - timedelta(days=3)).strftime(DATE_FORMAT)
+    date_from = (today - timedelta(days=settings.FIO_BANK_PROCESS_DAYS)).strftime(DATE_FORMAT)
```
ZoltanOnody: The tests will have to be refactored as well. 3 was a hard-coded constant, so if you change it they will fail. Sorry for that.

ricco386: no problem, I will fix it here

PR #246, opened 2016-11-30
245 CfP fixed edit url and prettified end date

mimmon:
edit url - replace event.uuid with talk.uuid
end_date prettified, in case it is not defined start of conference will be used instead
updated test to use cfp_end

### DIFF
--- a/konfera/event/views.py
+++ b/konfera/event/views.py
@@ -1,6 +1,6 @@
 import logging
 
-from datetime import timedelta
+from datetime import datetime, timedelta
 from smtplib import SMTPException
 
 from django import VERSION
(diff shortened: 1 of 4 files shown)


konfera/event/views.py
```diff
@@ -100,7 +100,10 @@ def post(self, *args, **kwargs):
             if settings.PROPOSAL_EMAIL_NOTIFY:
                 event_url = self.request.build_absolute_uri(reverse('event_details', args=[self.event.slug]))
                 edit_url = self.request.build_absolute_uri(reverse('event_cfp_edit_form',
-                                                                   args=[self.event.slug, self.event.uuid]))
+                                                                   args=[self.event.slug, talk_instance.uuid]))
+
+                end = getattr(self.event, 'cfp_end') or self.event.date_from
```
ricco386: this is not necessary:
A. either we get 404 when cfp is not allowed
B. event clean method should require cfp_end if cfp is allowed.

besides to let people know on the event start is way too late ;) I rather see exception and we know something is buggy (I already spot an issue 👍 #248, may be you did too and this was the reason for this code, but it should be handled elsewhere) than send wrong expectations to potential speakers. So I think use directly self.event.cfp_end

PR #247, opened 2016-11-30
Order update notification

ricco386:
I have create email notification when payment is processed. We want to let people know that order has been paid. In case of PayPal it is not that important, as the action is instant and user is redirected back to the order. For bank transfer this is important as user will be automatically notified that the payment has been received.

As a part of this I have added new field processing_fee that is added to order before going to paypal as we want to show the fee on the order so user is not confused that he has paid more than was the ticket price.


payments/utils.py
```diff
@@ -105,6 +110,51 @@ def _process_payment(order, payment, verbose=0):
         method=payment.get('payment_method', 'fiobank-transfer'),
     )
 
+    if settings.PAYMENT_PROCESS_EMAIL_NOTIFY:
+        event = order.event
+        template = EmailTemplate.objects.get(name='order_update_email')
+        subject = _('Your ticket for {event}.'.format(variable_symbol=order.variable_symbol,
```
ZoltanOnody: You don't need `variable_symbol`here.

payments/utils.py
```diff
@@ -105,6 +110,51 @@ def _process_payment(order, payment, verbose=0):
...
+                                                      event=event.title))
+
+        for ticket in order.ticket_set.all():
+            text_content = template.text_template.format(first_name=ticket.first_name,
+                                                         last_name=ticket.last_name,
+                                                         event=event.title,
```
ZoltanOnody: I'm not a big fan of formatting like this one. If we change the variable name from `text_content` for some reason, we will have to also indent all the lines bellow which will create a bunch of changes in git. (Just my opinion, feel free to ignore it.)

payments/views.py
```diff
@@ -103,6 +105,9 @@ def pay(request, order):
             order=order.pk, err=payment.error))
         messages.error(request, _('Something went wrong, try again later.'))
 
+        order.processing_fee -= paypal_fee
```
ZoltanOnody: Hmm.. this doesn't make sense to me.

Why are you adding the `paypal_fee` to `processing_fee` and then subtracting it?

konfera/templates/konfera/order_details.html
```diff
@@ -65,19 +65,61 @@ <h3 class="text-center panel-title">{% trans "Ordered tickets" %}</h3>
...
+                  <div class="row">
+                    <div class="col-xs-6">
+                        <strong>{% trans "Discount: " %}</strong>
+                    </div>
+                    <div class="col-xs-6">
+                        {{ order.discount|currency }}
+                    </div>
+                  </div>
+                  {% endif %}
+                  {% if order.processing_fee %}
```
ZoltanOnody: Processing fee is 0 by default. So it won't show up. Then the user clicks pay by paypal, it will increase and immediately decrease so it won't show up afterwards as well.

payments/utils.py
ricco386: yes, my fault I was experimenting with different subjects and forgot to remove. At the end I decided to use the same subject so in most clients the email end in the same thread as registration email which make sence.

payments/views.py
ricco386: we end up here only if there is problem with PayPal, otherwise he is redirected to PayPal and this code is never executed. If PayPal has error, than paypal_fee no longer applies as user have only chance to pay by bank transfer (or other method we implement in future)

PR #250, opened 2016-12-02
241 description and price info in registration form

mimmon:


mimmon:
the tests failed because of a bug in tests which are solved in my previous commit, after merging #249, it should work. I will make another PR later.

konfera/models/ticket.py
```diff
@@ -29,6 +29,7 @@ class Ticket(KonferaModel):
     last_name = models.CharField(max_length=128)
     email = models.EmailField()
     phone = models.CharField(max_length=64, blank=True)
+    # TODO isn't the description useless? The general description of the ticket provides type.description...
```
ricco386: Please remove the TODO. TODO should mark some future functionality, or change that shall be done due to some changes, you are asking question, that shall be either discussed in issue comments, or at mailinglinst, or chat.

To answer your question:
`ticket_type` has description that is displayed when user is registering the ticket.
`ticket` has description that is filled by user. We can call `register.form.RegistrationForm` without description or with it, as for tickets, such as press, or aid we would like to get some input from attendee before confirming the ticket. Seems description isnt best name :) but attendee describe something about himself so we can re-name to something more obvious... What about about_me?

konfera/templates/konfera/registration_form.html
```diff
@@ -1,10 +1,14 @@
...
+{% extends 'konfera/base.html' %}
+{% load i18n %}
+{% load custom_filters %}
 
 {% block content_top %}
     <h2>{% blocktrans %} Registration of a {{ type }} {% endblocktrans %}</h2>
 {% endblock %}
 
 {% block content_details %}
+    <p class="list-group-item-text">{{ description }}</p>
```
ricco386: Coul you please do the style same as on `/{event}/ticket` page? e.g. there is also ticket name, and everything is grouped in div `list-group`.

konfera/models/ticket.py
mimmon: sorry for this "todo" comment, I left it as a remark for myself and forgot to discuss properly. 
We can perfectly let it called description unless users complain.

PR #257, opened 2016-12-07
Added ability to select talk language

ricco386:


konfera/settings.py
```diff
@@ -10,6 +10,13 @@
 """
 CURRENCY = getattr(settings, 'CURRENCY', ('€', 'EUR'))
 
+TALK_LANGUAGE = getattr(settings, 'TALK_LANGUAGE',
+                        (
+                            ('SK', _('Slovak')),
```
dn0: Maybe the description for `SK` could be `Slovak/Czech`.

ZoltanOnody: They are 2 different languages so lets keep them separate.

konfera/migrations/0019_talk_language.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('konfera', '0018_auto_20161203_2305'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='talk',
+            name='language',
+            field=models.CharField(choices=[('SK', 'Slovak'), ('EN', 'English')], default='EN', max_length=2),
```
ZoltanOnody: I don't see Czech here. Maybe you created the migrations before you add the language to the list.

PR #258, opened 2016-12-08
Email notifier

ZoltanOnody:


konfera/settings.py
```diff
@@ -44,3 +44,9 @@
 Universal BCC for all notifications
 """
 EMAIL_NOTIFY_BCC = getattr(settings, 'EMAIL_NOTIFY_BCC', [])
+
+
+"""
+Number of days after the user is notified about unpaid order
+"""
+UNPAID_ORDER_NOTIFICATION_DAYS = getattr(settings, 'UNPAID_ORDER_NOTIFICATION_DAYS', 3)
```
ricco386: Lets change the default to 5 days.

I am thinking if we can make it, a little bit more flexible with minor effort. How about we add two more options:

* `UNPAID_ORDER_NOTIFICATION_REPEAT = 3  # How many times we should repeat the email notification`
* `UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY = 5  # delay in days`

konfera/models/order.py
```diff
@@ -32,6 +32,7 @@ class Order(KonferaModel):
     status = models.CharField(choices=ORDER_CHOICES, default=AWAITING, max_length=20)
     purchase_date = models.DateTimeField(auto_now_add=True)
     payment_date = models.DateTimeField(blank=True, null=True)
+    unpaid_notification_sent_at = models.DateField(blank=True, null=True)
```
ricco386: lets add `unpaid_notification_sent_amount` Int field.

payments/utils.py
```diff
@@ -173,3 +173,45 @@ def check_payments_status(verbose=0):
...
+
+
+def send_unpaid_order_email_notifications(verbose=0):
+    """
+    Send email to all users who has unpaid orders.
+    """
+    deadline = timezone.now() - timedelta(days=UNPAID_ORDER_NOTIFICATION_DAYS)
+
+    orders = Order.objects.filter(Q(status=Order.AWAITING) | Q(status=Order.PARTLY_PAID))
+    orders = orders.filter(date_created__lt=deadline, unpaid_notification_sent_at__isnull=True)
```
ricco386: Lets update order to fileter for `unpaid_notification_sent` set to NULL or `unpaid_notification_sent` is lass than `UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY` and `unpaid_notification_sent_amount` is less than `UNPAID_ORDER_NOTIFICATION_REPEAT`

payments/utils.py
```diff
@@ -173,3 +173,45 @@ def check_payments_status(verbose=0):
...
+                msg.send()
+            except (SMTPException, ConnectionRefusedError) as e:
+                logger.critical('Sending email raised an exception: %s', e)
+            else:
+                # increase count on email_template
+                template.add_count()
+                msg = 'Email template unpaid_ticket_notifier has been send to: %s' % ticket.email
+                logger.debug(msg)
+
+                order.unpaid_notification_sent_at = timezone.now()
```
ricco386: increase `unpaid_notification_sent_amount`

konfera/migrations/0019_unpaid_templates.py
```diff
@@ -0,0 +1,39 @@
...
+from django.db import migrations
+
+
+def forwards_func(apps, schema_editor):
+    EmailTemplate = apps.get_model("konfera", "EmailTemplate")
+    db_alias = schema_editor.connection.alias
+    text = """Dear {first_name} {last_name},\n\n
+order with your ticket for {event} hasn't been paid yet.\n\n
+Please pay for it as soon as possible otherwise your ticket will expire.\n\n\n
+
```
ricco386: Also we should send the link to order details, so user can review it.

PR #260, opened 2016-12-10
Updated contry name

ricco386:


mimmon:
This year Czechia became official English short name for Czech republic. So it was actually correct although it still sounds a bit strange after those 20 years of having no short form.

ricco386:
Really?

On 14 April 2016, the country's political leadership agreed to make Czechia the official short name. The new name was approved by the Czech cabinet on 2 May 2016.

So it seems this PR can be cancelled.

PR #263, opened 2016-12-15
Order pdf

ricco386:
Added ability to svate order as PDF, unified button looks on the project


konfera/event/views.py
```diff
@@ -224,34 +227,115 @@ def event_public_tickets(request, slug):
...
+    template_name = 'konfera/order_detail.html'
+    slug_field = 'uuid'
+    slug_url_kwarg = 'order_uuid'
 
-    if order.event:
-        update_event_context(order.event, context, show_sponsors=False)
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data()
+        context['order'] = self.object
+        context['allow_receipt_edit'] = True
```
ZoltanOnody: `context['allow_receipt_edit'] = self.object.status == Order.AWAITING`

and you can remove the condition bellow

konfera/event/views.py
```diff
@@ -224,34 +227,115 @@ def event_public_tickets(request, slug):
...
+class EventOrderDetailFormView(ModelFormMixin, EventOrderDetailView):
+    form_class = ReceiptForm
+
+    def dispatch(self, *args, **kwargs):
+        self.object = self.get_object()
 
-    if order.status == Order.AWAITING:
-        context['form'] = form = ReceiptForm(request.POST or None, instance=order.receipt_of)
+        if self.object.status != Order.AWAITING:
+            messages.error(self.request, _('You can not edit order. Please contact support.'))
```
ZoltanOnody: *You cannot edit your order. Please contact the support.*

sounds better in my opinion

konfera/event/views.py
```diff
@@ -224,34 +227,115 @@ def event_public_tickets(request, slug):
...
+
+    def post(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        form = self.get_form()
 
         if form.is_valid():
-            form.save()
-            messages.success(request, _('Your order details has been updated.'))
+            return self.form_valid(form)
+        else:
```
ZoltanOnody: don't use else after return

konfera/event/views.py
```diff
@@ -224,34 +227,115 @@ def event_public_tickets(request, slug):
...
+        if self.object.receipt_of.title:
+            context['display_receipt'] = True
 
-    return render(request=request, template_name='konfera/event/event_organizer.html', context=context)
+        return context
+
+    def render_to_response(self, context):
+        try:
+            response = PDFTemplateResponse(request=self.request,
+                                           template=self.template_name,
```
ZoltanOnody: This formatting looks so wrong IMHO.

ZoltanOnody:
I pulled the code, installed the requirements, clicked to "save as pdf" and got the following error.

image

[traceback](http://dpaste.com/0ZWTESP)

Any idea why?

ricco386:
@ZoltanOnody install the [wkhtmltopdf static binary](http://wkhtmltopdf.org/downloads.html).

ZoltanOnody:
OK. We have to mention is as dependency somewhere.

Btw, it seems to be little bit buggy.

image

[traceback](http://dpaste.com/2PTQXG4)

ricco386:
How did you manage to get that? does that binary works from command line? This seems like error from the binary not from the code...

I will add new option to enable PDF, disabled by default (so people does not get broken app) and add a documentation there. We should really write a proper documentation, this is the weakest place of the app :)

ZoltanOnody:
You are right. The binary doesn't work.

I think it's a good idea to have the PDF generation optional. :)

PR #265, opened 2016-12-18
259 email notifications

mimmon:
There is one fully prepared migration where we can put link to the order. Just wanted to make sure what should be in the link, is there any helper method or should I create it from regex in https://github.com/pyconsk/django-konfera/blob/master/payments/urls.py#L7  (something like `url = reverse('payment_options')` )?

There is also a function in payment utils that can be used as the next step - after several unsuccessful attempts no other email will be sent and order will be Cancelled - can be a separate Issue or I can do it in this issue.


konfera/settings.py
```diff
@@ -53,8 +53,12 @@
...
+How many times we should repeat the email notification
+"""
+UNPAID_ORDER_NOTIFICATION_REPEAT = 3
 
 """
-Number of days after the user is notified about unpaid order
+How long shoudl we wait to notify about missing payment
 """
-UNPAID_ORDER_NOTIFICATION_DAYS = getattr(settings, 'UNPAID_ORDER_NOTIFICATION_DAYS', 3)
+UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY = 5
```
ZoltanOnody: It has to be over-writable in settings.

konfera/settings.py
```diff
@@ -53,8 +53,12 @@
...
 EMAIL_NOTIFY_BCC = getattr(settings, 'EMAIL_NOTIFY_BCC', [])
 
+"""
+How many times we should repeat the email notification
+"""
+UNPAID_ORDER_NOTIFICATION_REPEAT = 3
 
 """
-Number of days after the user is notified about unpaid order
+How long shoudl we wait to notify about missing payment
```
ZoltanOnody: typo in *should*

konfera/settings.py
```diff
@@ -53,8 +53,12 @@
 """
 EMAIL_NOTIFY_BCC = getattr(settings, 'EMAIL_NOTIFY_BCC', [])
 
+"""
+How many times we should repeat the email notification
+"""
+UNPAID_ORDER_NOTIFICATION_REPEAT = 3
```
ZoltanOnody: It has to be over-writable in settings.

payments/utils.py
```diff
@@ -175,14 +176,20 @@ def check_payments_status(verbose=0):
             _process_payment(order, payment, verbose)
 
 
+def get_unpaid_orders(overdue=False):
+    deadline = timezone.now() - timedelta(days=UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY)
+    orders = Order.objects.filter(Q(status=Order.AWAITING) | Q(status=Order.PARTLY_PAID))
+    orders = orders.filter(Q(date_created__lt=deadline) | Q(unpaid_notification_sent_at__lt=deadline))
+    if overdue:
```
ZoltanOnody: What is `overdue` and why is it here if we don't use it anywhere?

payments/utils.py
```diff
@@ -175,14 +176,20 @@ def check_payments_status(verbose=0):
             _process_payment(order, payment, verbose)
 
 
+def get_unpaid_orders(overdue=False):
+    deadline = timezone.now() - timedelta(days=UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY)
+    orders = Order.objects.filter(Q(status=Order.AWAITING) | Q(status=Order.PARTLY_PAID))
+    orders = orders.filter(Q(date_created__lt=deadline) | Q(unpaid_notification_sent_at__lt=deadline))
```
ZoltanOnody: Hmm... I am not sure this query does what we want.

I think we need something like:
```
(Q(date_created__lt=deadline) & Q(unpaid_notification_sent_amount=0)) # we haven't sent any notification yet 
| Q(unpaid_notification_sent_at__lt=deadline)  # or we sent the notification X days ago.
```

mimmon: You're right, the original query will actually ignore the notification date.

payments/utils.py
mimmon: Overdue means it has not been paid within given period of time (i.e. after three notifications). So it is  just a preparation for cancelling unpaid orders.

payments/utils.py
```diff
@@ -175,14 +176,21 @@ def check_payments_status(verbose=0):
             _process_payment(order, payment, verbose)
 
 
+def get_unpaid_orders(overdue=False):
+    deadline = timezone.now() - timedelta(days=UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY)
+    orders = Order.objects.filter(Q(status=Order.AWAITING) | Q(status=Order.PARTLY_PAID))
+    orders = orders.filter(Q(unpaid_notification_sent_amount=0) & Q(date_created__lt=deadline) |
```
ZoltanOnody: I am not sure whether this means (A and B) or C  ...or... A and (B or C). Could you put there the parenthesis explicitly?

payments/utils.py
mimmon: I will do it, no problem. 
I also want to change the email template as requested in specs of the issue.

mimmon:
There is still some work to be done - the email template (in migration) should be changed to accept link to order and then add link in the code.

ZoltanOnody:
Okay, that's why the author of the PR should merge it. I just approved the changes because they seems to be OK. If there are any work in PR left, you can still commit them. :)

konfera/settings.py
```diff
@@ -70,3 +70,14 @@
 How long should we wait to notify about missing payment
 """
 UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY = getattr(settings, 'UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY', 5)
+
+"""
+Absolute url base with protocol
+"""
+ABSOLUTE_URL = getattr(settings, 'ABSOLUTE_URL', 'https://www.pycon.sk/')
```
ricco386: How about we rename it to SITE_URL, you can leave in comment that it is absolute url. I can imagine we will need SITE_NAME for the emails, possibly others as well.

konfera/settings.py
```diff
@@ -70,3 +70,14 @@
 How long should we wait to notify about missing payment
 """
 UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY = getattr(settings, 'UNPAID_ORDER_NOTIFICATION_REPEAT_DELAY', 5)
+
+"""
+Absolute url base with protocol
```
ricco386: please add a note it always have to have end without (once we use reverse) trailing slash (/) so it is explicit.

payments/utils.py
```diff
@@ -176,6 +176,12 @@ def check_payments_status(verbose=0):
             _process_payment(order, payment, verbose)
 
 
+def get_full_order_url(order):
+    ABSOLUTE_URL = settings.get('ABSOLUTE_URL')
+    full_url = '{}order/{}'.format(ABSOLUTE_URL, order.uuid)
```
ricco386: please use [reverse](https://docs.djangoproject.com/en/1.10/ref/urlresolvers/#reverse) this wont work if we change url for some reason...

konfera/settings.py
mimmon: I will add it. I thought about testing it in code just for sure, but it is probably cheaper to get error soon and have it fixed than check every single time if it is in required format. Am I right?

PR #268, opened 2016-12-22
256 email util

mimmon:
initial commit, there are some questions/ambiguities to be decided


konfera/utils.py
```diff
@@ -41,3 +49,48 @@ def update_event_context(event, context, show_sponsors=True):
...
+        logger.critical('Invalid text template (required) for the input {}.'.format(text_content))
+        # raise EmailTemplateError("Email template is not valid for the input.")
+    if not html_content:
+        logger.warning('Invalid html template (not required) for the input {}.'.format(html_content))
+
+    to = addresses.get('to', [])
+    cc = addresses.get('cc', [])
+    bcc = addresses.get('bcc', EMAIL_NOTIFY_BCC)
+
+    # if text template does not match - do we want to raid the exception or to send unformatted template?
```
ricco386: I think exception should be here, because:

If there is an error in template first who encounters it will raise exception, and admin should amend accordingly (he has a chance to fix after first encounter). If he does not exceptions will always be raised and user does not recaive email, on the other hand if we always send email with wrong text, user will not receive any useful info (to me it is almost the same as not receive email at all), and admin will not see this in logs, monitoring, anywhere so he doesnt have the chance to fix. And in this case fix is re-wording the template in DB.

konfera/utils.py
```diff
@@ -41,3 +49,48 @@ def update_event_context(event, context, show_sponsors=True):
...
+    if required_keys.issubset(set(input.keys())):
+        return raw_template.format(**input)
+    return
+
+
+class EmailTemplateError(Exception):
+    pass
+
+
+def send_email(addresses, subject, template, input=None, verbose=0):
```
ricco386: input is built in function, please consider different name.

we are passing also the content of the templates, how about we just pass the name (that should be called in the DB) and we will do a search in the DB here. You can have separete search_template function that will raise error if not found or return both text and html template if exists in DB. I can imagine it would clear the code elsewhere. Just a thing to consider, havent look at the place where we do call them, this is just a thought.

konfera/event/views.py
```diff
@@ -339,7 +339,7 @@ def render_to_response(self, context):
                                            context=self.get_context_data(), show_content_in_browser=False,
                                            cmd_options={'margin-top': 10, 'zoom': 1, 'viewport-size': '1366 x 513',
                                                         'javascript-delay': 1000, 'footer-center': '[page]/[topage]',
-                                                        'no-stop-slow-scripts': True, 'dpi': 130, 'zoom': 1.33})
```
ZoltanOnody: Are you sure the right value is 1?

```
>>> d = {'a': 1, 'b':2, 'a':3}
>>> d['a']
3
```

So I think (not to change the functionality) we should leave the `1.33` and remove the `1`. @ricco386  ?

konfera/utils.py
```diff
@@ -70,6 +70,7 @@ def validate_email_template(raw_template, formatting_dict, required=True):
                                      'required:{} !~ formatting:{}'.format(required_keys, set(formatting_dict)))
         else:
             logger.warning('Not all required fields of the template were found in formatting dictionary.')
+            return raw_template
```
mimmon: Is there a possibility that we want unformatted/incomplete template to be shown in specific cases or can we remove this 'else' (and required parameter for this function? The other option is to convert the known parameters and unknown left untouched.

konfera/register/views.py
```diff
@@ -47,31 +46,16 @@ def _register_ticket(request, event, ticket_type):
...
-                messages.success(request, _('Thank you for ordering ticket. You\'ll receive confirmation email soon.'))
-
+            formatting_dict = {
+                'first_name': new_ticket.first_name,
+                'last_name': new_ticket.last_name,
+                'event': event.title,
+                'order_url': order_url,
+                'event_url': event_url,
+            }
+            addresses = {'to': [new_ticket.email], 'bcc': settings.EMAIL_NOTIFY_BCC}
```
ricco386: we do already set bcc in utils, so this is duplication, please remove...

konfera/utils.py
```diff
@@ -1,8 +1,18 @@
...
 from decimal import Decimal, ROUND_UP
+from smtplib import SMTPException
 
+from django.core.mail import EmailMultiAlternatives
+
+from konfera.models.email_template import EmailTemplate
 from konfera.models.sponsor import Sponsor
 from konfera.models.order import Order
 from konfera.settings import GOOGLE_ANALYTICS, NAVIGATION_ENABLED, NAVIGATION_URL, NAVIGATION_LOGO, NAVIGATION_BRAND
+from konfera.settings import EMAIL_NOTIFY_BCC
```
ricco386: Why do you need to import twice from the same file, for multiline imports use brackets...

PR #273, opened 2016-12-22
Update PDF not to generate navigation and footer

ricco386:


ricco386:
Sorry, fixed in master.

PR #292, opened 2017-01-22
Attendee checkin

### LINKED ISSUE #289: Add view for checking attendees in.
Frontend view, that is accessible for volunteers.

Volunteers (logged in user, not necessarily admin account) should be able to search tickets. It would display ticket details, and volunteer can confirm checking-in (eg. post button will change status).

You can add ability to search by multiple fields (lastname, email, ordernumber, etc...).

ZoltanOnody:
closes #289 

image

image

todo:
- write tests

PS: the *documentation* is (probably) in a wrong format so it will need to be changed (I don't know how).


konfera/event/views.py
```diff
@@ -353,3 +350,69 @@ def render_to_response(self, context):
...
+        if search:
+            queryset = queryset.filter(
+                Q(first_name__icontains=search) |
+                Q(last_name__icontains=search) |
+                Q(email__icontains=search)
+            )
+
+        requested_only = self.request.GET.get('requested_only')
+        if requested_only:
+            queryset = queryset.filter(status=Ticket.REQUESTED)
```
ricco386: How about we add Registered checkbox? Or change this one.

Original idea behind ticket statuses is this:
* requested - user just fills the order, or requested donation for the ticket but hasnt paid yet or donation hasnt been approved yet (i can imagine due to human error this people show up at registration).
* registered - user has paid for the ticket or the donation has been approved
* checked-in - user is on the premise of the conference
* cancelled - user either has not paid after some time, or paid and has canecelled and we have returned the payment.

ZoltanOnody:
> do we want to search by order number (variable symbol) - low priority just thinking if it might help at the registration desk.

I think it won't be necessary but I'd certainly show the variable symbol in the detail view because 2 attendees can have the same name.

>  pagination number drop-down 20, 50, 100, 500, 1000 - low priority (i can imagine that volunteer at registration display all attendees) and in new tab changes the status, not sure if useful?

I think 50 is enough it's 1/8 of our attendees and if the volunteer will use the search it will certainly fit to one page.

Other suggestion have been implemented.

PR #293, opened 2017-01-23
Add django-dynamic-form to the project

ZoltanOnody:
todo:
- [ ] regenerate the migrations before merging the PR
- [ ] write tests


questionnaires/views.py
```diff
@@ -0,0 +1,32 @@
...
+
+from dynamic_forms.models import FormModel
+from dynamic_forms.views import DynamicFormView
+
+from konfera.models import Ticket
+
+from .models import FormForTicket
+
+
+class ShowFormForTicket(DynamicFormView):
```
ZoltanOnody: @ricco386 I am having some difficulties with this class based view.

when I visit `/form/1/ticket/my_uuid/` the `dispatch` is invoked as supposed but when I submit the form, `form_valid` is not invoked, however the `form_valid` in `DynamicFormView` is. So basically I cannot override the method and I have no idea why. Any suggestions?

ZoltanOnody: In the meantime I found out where is the problem.

questionnaires/requirements.txt
```diff
@@ -0,0 +1 @@
+django-dynamic-forms-0.5.3
```
ricco386: please fix the typo

requirements.txt
```diff
@@ -4,3 +4,4 @@ django-sitetree==1.6.0
 mock==2.0.0
 pytz==2016.7
 Pillow==3.4.2
+django-dynamic-forms-0.5.3
```
ricco386: please fix the typo

ZoltanOnody:
As the app is going to be refactored and simplified (#320), I close this PR.

PR #294, opened 2017-01-29
Added test if title has been filled in, so admin does not throw exception

ricco386:
Type KeyError
Value ''
Module builtins
Location konfera/models/ticket.py, __str__:37

### DIFF
--- a/konfera/models/ticket.py
+++ b/konfera/models/ticket.py
@@ -33,11 +33,19 @@ class Ticket(KonferaModel):
     order = models.ForeignKey('Order', blank=True)
 
     def __str__(self):
-        return '{title} {first_name} {last_name}'.format(
-            title=dict(Speaker.TITLE_CHOICES)[self.title],
-            first_name=self.first_name,
-            last_name=self.last_name
-        ).strip()
+        if self.title:
+            full_name = '{title} {first_name} {last_name}'.format(
+                title=dict(Speaker.TITLE_CHOICES)[self.title],
+                first_name=self.first_name,
+                last_name=self.last_name
+            )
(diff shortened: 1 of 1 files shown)


konfera/models/ticket.py
```diff
@@ -33,11 +33,19 @@ class Ticket(KonferaModel):
     order = models.ForeignKey('Order', blank=True)
 
     def __str__(self):
-        return '{title} {first_name} {last_name}'.format(
-            title=dict(Speaker.TITLE_CHOICES)[self.title],
-            first_name=self.first_name,
-            last_name=self.last_name
-        ).strip()
+        if self.title:
```
ZoltanOnody: This seems too complicated. What about 

```
return '{title} {first_name} {last_name}'.format(
    title=dict(Speaker.TITLE_CHOICES).get(self.title, ''),
    first_name=self.first_name,
    last_name=self.last_name
).strip()
``

File: konfera/models/ticket.py [hunk 2]
```diff
@@ -33,19 +33,11 @@ class Ticket(KonferaModel):
...
 
     def __str__(self):
-        if self.title:
-            full_name = '{title} {first_name} {last_name}'.format(
-                title=dict(Speaker.TITLE_CHOICES)[self.title],
+        return '{title} {first_name} {last_name}'.format(
+                title=dict(Speaker.TITLE_CHOICES).get(self.title, ''),
                 first_name=self.first_name,
                 last_name=self.last_name
             )
```
ZoltanOnody: we need to `.strip()` it because if there is not title the text will start with a space

PR #297, opened 2017-01-31
Discount code error

ricco386:
Added ability to have discount code at registration form as a new field. Updated exception in admin area that was causing the system to crash.


konfera/register/forms.py
```diff
@@ -1,21 +1,38 @@
...
 
         if description_required:
             self.fields['description'] = forms.CharField(
                 widget=forms.Textarea({'placeholder': _('Please tell us something about yourself...')}),
                 required=True, label=_('Description'))
+
+    def clean_discount_code(self):
+        data = self.cleaned_data['discount_code']
+
+        if data:
```
ZoltanOnody: The code is nested so I don't like it. :-D 

what about
```
if not data:
    return None

# the rest of the validation
```

PR #321, opened 2017-08-18
Top level project cleanup

ZoltanOnody:


ricco386:
You have crated the PR against master, we have agreed that until we have minimal working version we will have it in a branch and than switch it to the master.

Can you change it to be against branch  v2.0? Otherwise looks good, we have deleted a lot of stuff ;)

PR #324, opened 2017-08-19
Remove views and templates from project

ZoltanOnody:


ricco386:
Looks OK, even with the previous PR merges it ended with conflict. I am wondering why we left the forms in, are we going to re-use them? Since we will do the rest API it seems there would be no use for this objects anyway and since nothing seems to be calling them I suggest let's remove them, or you want to do it in separate PR once we are sure?

Issue #320, opened 2017-08-18
Refactor!


ricco386:
Get rid of the "cms" stuff in th models and do a cleanups

PR #325, opened 2017-08-21
Removed unecessary models, files and roleaded files due to issue #320

ricco386:


ZoltanOnody:
fix the test, otherwise LGTM.

ricco386:
I have reviewed, refactored all the konfera tests. They use model-mommy so they are more readable and flexible once we introduce new changes to the structure. I have disabled payments test, however, they were only for views and utils, since we are going to drop all views and build just the rest API it probably doesn't matter anyway.

ZoltanOnody (approved):

PR #327, opened 2017-09-21
Added flag for event to be able to add grouping for talks

ricco386:
Once we add a same flag for certian talks we can mark them as track, base filters etc...


konfera/models/talk.py
```diff
@@ -51,6 +51,7 @@ class Talk(KonferaModel):
         blank=True,
         null=True,
     )
+    title = models.CharField(max_length=32, blank=True)
```
ZoltanOnody: I believe you meant `flag` and not `title`. This causes a collision.

ZoltanOnody (changes requested):
see above

konfera/models/talk.py
```diff
@@ -51,6 +51,7 @@ class Talk(KonferaModel):
         blank=True,
         null=True,
     )
+    flag = models.CharField(max_length=32, blank=True)
```
ZoltanOnody, line 54: what about `null=True`? (maybe it's not needed, just making sure)

ricco386:
Yes, you are right. Thanks for spotting it.

konfera/models/talk.py
ricco386: In all other models for text fields, we use just blank=True, so I think this is fine since we are consistent with other models.

ZoltanOnody (approved):
LGTM

PR #329, opened 2017-11-07
V2.0 aid

ricco386:


ZoltanOnody:
tests are failing please fix them

konfera/models/ticket_type.py
```diff
@@ -86,9 +87,23 @@ def status(self):
     def issued_tickets(self):
         return self.ticket_set.exclude(status=Ticket.CANCELLED).count()
 
+    def calculate_available_tickets(self, numeric_only=False, negative_usage=False):
+        if self.usage == -1:
+            if numeric_only:
+                left = 32767
```
ZoltanOnody: where did this number came from?

ZoltanOnody: why don't we use `float('inf')` ?

konfera/models/ticket_type.py
```diff
@@ -86,9 +87,23 @@ def status(self):
     def issued_tickets(self):
         return self.ticket_set.exclude(status=Ticket.CANCELLED).count()
 
+    def calculate_available_tickets(self, numeric_only=False, negative_usage=False):
```
ZoltanOnody: I don't like the structor of this method, my personal preference would be to add there more `return` this way I have to look on the whole method to see that the value of `left` doesn't change anymore.

ZoltanOnody (changes requested):
otherwise LGTM

konfera/models/ticket_type.py
ricco386: Well, I think it really depends, in theory, we should minimize the number of returns in each routine. It's harder to understand a routine if, reading it at the bottom, you're unaware of the possibility that it returned somewhere above.

Use a return when it enhances readability. In certain routines, once you know the answer, you want to return it to the calling routine immediately. If the routine is defined in such a way that it doesn't require any cleanup, not returning immediately means that you have to write more code.

I have tried to rewrite the function to use less if's however we do not return 0 if the value is negative somehow, but I guess it make sence if available tickets is -7 than you can no longer sell ;)

ricco386:
Unitests are (were always) OK, codecov is failing as it doesnt have version compare agains (not sure why).

ZoltanOnody (approved):

